PR #341, opened 2014-06-10
Detect and report hashbag as unsupported

ozten:
If I run `sjs` over a NodeJS source which has `#!/usr/bin/env node` then I get the error

    Error: Line 1: Invalid regular expression

I wasn't able to find a clean way to remove the offending line from the source and continue parsing.

Perhaps throwing a more direct error message? This PR would instead give the error

    Error: Line 1: HashBang interpreter directive is not supported


disnet:

PR #360, opened 2014-07-02
Magic vars for repeaters

natefaubion:
This adds a few magic pattern vars that can only be referenced in repeaters: `$$index`, `$$index1` (plus 1), and `$$length`.


disnet:

PR #405, opened 2014-10-16
Implementation of Syntax-param in sweet.js

VimalKumarS:
Hello,

File updated - Parser.js / Pattern.js
File added - Test1/Test2/test3.js and Debug-syntax.js
Syntax param example

```js
macro aif {
    case {
        $aif_name  
        ($cond ...) {$tru ...} else { $els ... }

    } => {
     SyntaxParameter(it, $cond ... , aif ,
        return #
	  {
            (function () {
                 if ($cond ...) {
                    $tru ...
                } else {
                    $els ...
                }
            })
	
        })
    }
}



macro unless {
    case { 
	  $unless_name 
	  ($cond ...) { $body ...} } => {
 return #{
        while (true) {
            aif ($cond ...) {
                // `it` is correctly bound by `aif`
                console.log("loop finished at: " + it);
                            } else {
                $body ...
            }
        }}
    }
}

x=2
unless (x) {
    console.log(it)
}
```

Please review once, let me know if this break the current working, or if have any example that fails.

thanks
Vimal


toothbrush7777777:
@VimalKumarS  Any update on this?

VimalKumarS:
I am working on this, with different approach or binding the syntax param as part of compiled macro and later binding it during macro expansion, as in scheme..

Issue #174, opened 2014-01-04
empty group repeat fails


disnet:
```js
let m = macro {
	rule { { $(case ($x ...)) ... } } => {
		$([$x (,) ...]) (;) ...
	}
}

m {

}
```

Throws an:

    Error: ASSERT: must have a least one non-scalar in repeat

disnet:
I'm mass closing issues due to the release of 1.0 obviating lots of bugs. If you feel this issue is something we still need to address feel free to reopen.

PR #440, opened 2014-12-28
Fixes issue #174. Instead of failing hard on empty repeats, bail early.

trxcllnt:


disnet:
Ah, right. When does this actually trigger? Can you include a test case?

disnet:

PR #445, opened 2015-01-08
Highlight macro expansions in editor

levjj:
Added a checkbox to the editor which allows highlighting of macro names and expansions.  Source Maps are used to map areas in the generated code back to the macro that generated it.

In order to be less obtrusive, the macro highlighting is only shown when the cursor in the editor is at the corresponding macro name.

highlighting


src/sweet.js
```diff
@@ -208,16 +208,15 @@
...
-        }
-        return {
-            code: codegen.generate(ast, _.extend({
+            result.code = output.code;
+            result.sourceMap = output.map.toString();
+        } else {
+            result.code = codegen.generate(ast, _.extend({
                 comment: true
             }, options.escodegen))
         };
```
disnet: broke the build because jshint doesn't like this unnecessary `;` :)

disnet:
If I try and use the class macro from the homepage some of syntax is not getting highlighted. Any idea why not?

edit_sweet_js

levjj:
Thanks for testing this out!  There was a bug in the actual CodeMirror highlighter that skipped certain parts of the text.  Fortunately, I was able to fix this with a one-line change.

disnet:
Thanks!

PR #451, opened 2015-02-05
Recognize function declarations

jugglinmike:
Currently, the `:expr` pattern matches function declarations as well as
function statements. Although I thought the fix would be simple, the underlying
bug is more far-reaching than I anticipated. I'd very much like to contribute
to this project, but I could really use a pointer or two in order to proceed.

Would anyone mind looking this over and offering suggestions?


disnet:
Yeah this is a bit trickier than it might first appear.

Since FunctionDeclaration and FunctionExpression can look the same when they are both named you have to use the parsing/enforestation context to tell them apart. The standard way that say esprima does this is by writing a normal recursive descent parser with a function per grammar production.

A bit of a refactor but probably what we should do.

disnet:
Handled by #485

PR #452, opened 2015-02-06
Fix issues with readtables

ricochet1k:
I've been making modifications to the jsx-reader project and I have been finding lots of problems with readtables, especially when they are nested in tokens read by other readtable extensions. I think I found the solution to one problem, but there are still other things going on.

The fix commit will make the startRange and endRange lines work, but the last expect is still broken. The parser is not able to parse its own output?

Another issue is "Invalid mapping" errors when generating sourceMaps, with a negative column number: `"original":{"line":194,"column":-91}` . It comes from generating tokens with a {start:} that is earlier than the reader's current line. There doesn't appear to be a better way to record the source location. Ideally one could pass in a token, or get a more complete location state from the reader, something like `reader.getLocation()` instead of just `reader.index`. I am not sure what an appropriate fix would be.

What is the best way to move forward with this? Readtables are a great feature, but the weak API is difficult to work with.


disnet:

Issue #437, opened 2014-12-21
proposed changes to syntax binding forms


disnet:
So right now we have some problems when it comes to our binding forms for macros. At the moment we have two, the recursive **declaration** form and non-recursive **let** form:

```js
// declaration form
macro m {
    ...
}

// let form
let m = macro {
    ...
}
```

They are modeled after the definition forms in scheme:

```scheme
(define-syntax m
  ...)

(let-syntax [(m ...)]
  ...)
```

The problem is that these are not a good analog for JS forms (`var`, `let` and `function` declarations). In particular let macros are not really a perfect analogue to `let` (temporal dead zone etc.). It also doesn't give us a nice general syntax binding form for things other than macros.

So, here's my idea. We add a new binding form, `stx`, as a drop-in replacement for `let` on compiletime values and remove the let macro form.

```js
stx m = ...
```

In particular, its behavior is:

1.  duplicate declarations in the same scope are an error
2.  block scoping so no hoisting
3.  temporal deadzone, references before `stx` are unbound except in nested structures
4.  `m` is not bound in the initializer (not recursive by default)

```js
// puts a number into the compiletime environment
stx m = 42;
// puts a normal function into the compiletime environment
stx m = function(stx) { ... }
// puts the primitive form of a macro onto the compiletime environment
stx m = macro { function(stx) { ... } }
// macro is a macro
stx m = macro { rule {} => { 42 } }
stx m = macro { case {_} => { #{42} } }
// and can be recursive by naming itself
stx m = macro m { rule {} => { m } }
stx m = macro m { case {_} => { #{m} } }

// equiv of define-syntax-rules
stx m = macro rule {} => { 42 }
stx m = macro case {_} => { #{42} }
// and the recursive form
stx m = macro m rule {} => { m }
stx m = macro m case {_} => { #{m} }
```

We could also use `syntax` instead of `stx` (though `stx` has a nice symmetry with `var` and `let` in terms of length). Perhaps other variants?

And just like function declarations we can also have a macro declaration form:

```js
macro m {
    rule {} => {}
}
```

That expands to `stx m = ...`.

Does this seem reasonable to people?

disnet:
I'm mass closing issues due to the release of 1.0 obviating lots of bugs. If you feel this issue is something we still need to address feel free to reopen.

PR #461, opened 2015-03-16
Scope hygiene

### LINKED ISSUE #430: "match" hygiene mismatch inside nested cases
If a case macro is created in the body of an outer case macro, the inner cases' `syntax` invocations are renaming their `match` identifiers to an undefined reference. I imagine this is because `syntax` breaks hygiene to capture `match` inside `syntaxCase`.

Here's my example:

```javascript
macroclass inv { pattern { rule { .$fn($args ...) } } }
macro extends_macro {
	case infix { $Macro | _ $Parent { $rules ... } } => {
		var rules = #{ $rules ... }.concat(#{
			case { _ $inv:inv  $[...] ; } => {
				console.clear();
				console.log("breaks after this");
				return #{
					$Parent .$inv$fn($inv$args $[...]);
				}
			}
		});
		return withSyntax($macro_rules ... = rules) #{
			macro $Macro { $macro_rules ... }
		}
	}
}

macro Base {
	rule { .foo(); } => { console.log("foo"); }
}

Impl extends_macro Base {
	case { _ .bar(); } => { return #{ console.log("bar"); } }
}

Impl.foo().bar();
```

http://sweetjs.org/browser/editor.html#macroclass%20inv%20%7B%20pattern%20%7B%20rule%20%7B%20.$fn($args%20...)%20%7D%20%7D%20%7D%0Amacro%20extends_macro%20%7B%0A%09case%20infix%20%7B%20$Macro%20%7C%20_%20$Parent%20%7B%20$rules%20...%20%7D%20%7D%20=%3E%20%7B%0A%09%09var%20rules%20=%20#%7B%20$rules%20...%20%7D.concat(#%7B%0A%09%09%09case%20%7B%20_%20$inv:inv%20%20$%5B...%5D%20;%20%7D%20=%3E%20%7B%0A%09%09%09%09console.clear();%0A%09%09%09%09console.log(%22breaks%20after%20this%22);%0A%09%09%09%09return%20#%7B%0A%09%09%09%09%09$Parent%20.$inv$fn($inv$args%20$%5B...%5D);%0A%09%09%09%09%7D%0A%09%09%09%7D%0A%09%09%7D);%0A%09%09return%20withSyntax($macro_rules%20...%20=%20rules)%20#%7B%0A%09%09%09macro%20$Macro%20%7B%20$macro_rules%20...%20%7D%0A%09%09%7D%0A%09%7D%0A%7D%0A%0Amacro%20Base%20%7B%0A%09rule%20%7B%20.foo();%20%7D%20=%3E%20%7B%20console.log(%22foo%22);%20%7D%0A%7D%0A%0AImpl%20extends_macro%20Base%20%7B%0A%09case%20%7B%20_%20.bar();%20%7D%20=%3E%20%7B%20return%20#%7B%20console.log(%22bar%22);%20%7D%20%7D%0A%7D%0A%0AImpl.foo().bar();

disnet:
# Massive Hygiene Simplification

So @mflatt of Racket fame has recently been working on a simplified hygiene algorithm for Racket that uses *scope sets* rather than marks and renames. His notes outlining the approach are [here](http://www.cs.utah.edu/~mflatt/scope-sets-2/). It's crazy good.

Rather than do what I really should be doing (writing my thesis) I've hacked it up in sweet.js.

Previously sweet.js was using the standard Scheme/Racket approach of marks/renames and definition contexts. This was always hard to understand how it worked and we had a number of bugs that I could never track down. `resolve` was also super slow.

The basic idea of scope sets is that every syntax object is marked with the set of scopes it is in and each binding form (e.g. `function`) creates a new scope:

```js
function foo(x^{a}) {
    function bar(x^{a, b}) {
        return x^{a, b};
    }
}
```

There is a global map that associates each identifier and scopeset with a new binding so the above might have a map like:

```js
{
    x: [{
        scopeSet: [a],
        binding: 1
    }, {
        scopeSet: [a, b],
        binding: 2
    }]
}
```

and fully expanding it with `resolve` would then give

```js
function foo(x1) {
    function bar(x2) {
        return x2;
    }
}
```

There are a few more details about macro invocation, recursive macro definitions, and `let`/`var` but that is the basic idea.

The big win here is that hygiene is mostly understandable now (check out how tiny [`resolve`](https://github.com/mozilla/sweet.js/blob/scope-hygiene/src/stx/resolve.js#L18-L44) is now!) and faster (even without the memoization we were doing in the old `resolve`).

A few other things snuck into this PR.

# New names for macro definitions

The current `let name = macro {}` is bad. I proposed some new syntax ideas in #437 and this is the start of implementing them. Internally there is now only one macro definition form `stxrec` which as the name suggests is recursive. On a suggestion from @mflatt the non recursive form is then built from the recursive form (implementation is [here](https://github.com/mozilla/sweet.js/blob/scope-hygiene/macros/stxcase.js#L7-L54)).

So right now the forms look like:

```js
stxrec m {
    rule {} => {}
}
stxnonrec m {
    rule {} => {}
}
```

The `stxnonrec` is obviously a terrible name and just a placeholder. I think we want:

```js
stxrec m = macro {
    rule {} => {}
}
stx m = macro {
    rule {} => {}
}
// using the `x = macro {}` is suggestive with normal `let`/`var`
// and also allows us to smoothly use the same syntax
// to put things other than macros into the compiletime env
stx x = 42;
stx x = function(x) { return x; };
// with maybe a convenient macro declaration form
macro m {
    rule {} => {}
}
```

# Macro definitions no longer break hygiene to capture `match`

Probably fixes #430 and others.

So sorry I ever did that :(

`macroclass` currently doesn't work because it was breaking hygiene and I haven't updated it yet.

# localExpand takes a stop list

At the moment it works by matching the name and the next delimiter token.

# Phasing

Racket has `begin-for-syntax` that let's you define stuff for use inside a macro without going to the trouble of pulling it out into a separate module and importing it for syntax. I've added `forPhase` to do the same:

```js
forPhase 1 {
    stxrec id {
        rule { $x } => { $x }
    }
}
stxrec m {
    case {_ } => {
        var x = id 42;
        return [makeValue(x, #{here})];
    }
}
```

There's probably some cleaner syntax we could come up with.

You can import into an arbitrary phase now:

```
import { m } from "mod.js" for phase 3;
```


vendethiel:
That sounds absolutely amazing :-).

natefaubion:
This is looking great! Bikeshed time.

Maybe instead of `forPhase` we could have a declaration sort of like `export var` or `export function`.
```js
phase 1 var foo = 42;
phase 2 function foo() { return 42 }
```

It also mirrors the `import` syntax. I personally dislike nesting declarations in `{}` that can be used outside of it, because it feels like it breaks lexical scope. But that may just be me.

I also don't really like `stx` for a keyword, especially since we commonly use it as an identifier in macros. Could we just use the full word?

disnet:
Bikesheds are the best!

Agreed `forPhase` indentation is the worst.

Some thoughts on `phase`:

```js
// good but...
phase 1 var foo = 42

// we have a convenient for phase 1 import form
import fs from 'fs' for macros

// kinda odd
macros var foo = 42
// also odd
for macros var foo = 42
// definitely odd
macros macro m {
    ...
}

// crazy?
var foo for phase 1 = 42
var foo for macros = 42
macro m for macros {
    ...
}
macro m for phase 1 {
    ...
}
macro m phase 1 {
    ...
}
var foo @ phase 1 = 42, bar @ phase 3 = 100;
var foo @ macros = 42
var @ phase 1 foo = 42;
var @ macros foo = 42; 
```

Yeah I've gone back and forth on `stx` vs `syntax`. `stx` is nice since it has the same length as `var` and `let`. Also `stxrec` is nice but `syntaxrec` feels pretty long. Not really sure where I'm landing on this one right now.

```js
stx foo = 42;
stxrec foo = 42;

syntax foo = 42;
syntaxrec foo = 42;
```

disnet:
Another `phase` thought, if we go with the `phase` as `export` form then `macros` could be an alias to `1`. Makes it a bit more declarative.

```js
phase macros var foo = 42;
phase templates var foo = 42;
// maybe not so bad?
phase macros macro m {
   ...
}
phase templates macro m {
  ...
}
```

natefaubion:
You can just lose the `s`.

```js
import foo from 'foo' for macro;
import goo from 'goo' for template;
import hoo from 'hoo' for phase 5;

macro foo {
}

macro var bar = 42;
macro function baz() {
}

template var a = 12;

phase 5 var b = 88;
```

disnet:
But then to put a macro declaration in phase 1 you have `macro macro m { }`.

disnet:
Or maybe that's just not allowed? Have `macro stx foo = macro {}` (or `syntax` if we go with that) but no standalone `macro m {}` form?

disnet:
Superseded by #485

PR #465, opened 2015-03-26
Automatic Macro Refactoring (Macrofication)

levjj:
For rule macros, it is possible to automatically refactor code with macros by using the template (body of the rule macro) to match syntax and replace it with the corresponding macro invocation based on the pattern.  This pull request adds support for this type of "macrofication" refactoring.  In order to use this features, I added a command line flag (-r) and integrated the refactoring into the sweet.js editor.  Using the editor is definitely the preferred way of refactoring as it highlights the code and let's you chose which code to replace by a macro:

letall

The macrofication option in the editor is off by default but seems to work pretty good.  Limitations are some special kind of "expr" pattern classes and locally scoped macro definitions.  Also, the highlighting is mutually exclusive with the expansion highlighting due to a limitation in CodeMirror.  The only way around this problem would be to write some extension to CodeMirror that supports multiple syntax highlighters at the same time.

This pull request is based on the patch for supporting repeated variables: [#540](https://github.com/mozilla/sweet.js/pull/450).

### DIFF
--- a/browser/scripts/reverse.js
+++ b/browser/scripts/reverse.js
@@ -25,12 +25,11 @@
 // DONE
 // support simple macro classes
 // get replacement
-//
-// TODO
 // check whether replacement compiles
 // do not match own macro definition
-// coalesce replacements
-// replace all
+//
+// TODO
+// command line interface
 // editor integration
 //
 // Future:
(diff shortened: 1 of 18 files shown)


src/reverse.js
```diff
@@ -0,0 +1,287 @@
...
+  THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+// DONE
+// support simple macro classes
+// get replacement
+// check whether replacement compiles
+// do not match own macro definition
+//
+// TODO
```
disnet: these are done right? maybe cleanup these comments

browser/scripts/reverse.js
```diff
@@ -0,0 +1,279 @@
...
+                    var idx = path[path.length - 1];
+                    var stx = path[path.length - 2];
+                    stx.splice(idx + 1, 2);
+                }
+            }
+        }, this.pattern);
+    };
+    function startRange(token) {
+        if (!token)
+            return 9999999999;
```
disnet: magic number. should this just throw an assert failure if `!token`?

PR #485, opened 2015-08-02
Redesign

disnet:
# Redesign

So I've been working on a rewrite/redesign of sweet.js. Turns out that what we currently have is kinda a rolling collection of hacks that are in desperate need of some rethinking. Things are slow and just getting slower in the branch with ES6 module support so something has to give.

What I'm pushing right now is definitely a work in progress but I think it shows promise. In particular I think it's much more comprehensible for people coming to the codebase for the first time. Read on if you're interested in helping out or just curious as to what might be changing.

Comments/opinions requested.

The current todo status looks something like this:

- [x] add support for all forms (right now I have some of the obvious ones but `switch`, `yield` and lots more aren't supported)
- [x] add hygiene (actually should be straightforward, hooks are already in the right places)
- [ ] add declarative macros (`rule` and `case`, currently just primitive macros "work")
- [ ] add module support (not straightforward but doable)
- [ ] add infix macros
- [ ] add custom operators
- [ ] spec out multi-token equivalent
- [ ] add line number and sourcemap support
- [ ] test each syntactic form
- [ ] port old tests
- [ ] add perf benchmarks
- [ ] add descriptive error messages

## No more destructuring

One of the big areas of slowdown came from the fact that we weren't doing real parsing. Expansion worked by building up a partial AST (`TermTree`) and then throwing all that work away by destructuring the partial AST back to an array of tokens and fed that to esprima to actually produce a real AST.

Now instead of doing all the parsing work twice, we just build the complete AST. This is being handled by two data structures, a `Term` (roughly equivalent to the current `TermTree`) that acts as a partial AST (some terms hold syntax objects and some hold other terms) and a `Node` that is just an [ESTree](https://github.com/estree/estree) representing the complete AST.

A `Term` has two methods, `parse` and `expand`. The `parse` method returns a new corresponding `Node` while `expand` is roughly equivalent to `expandToTermTree` in the current expander (ie handles some hygiene details and walks down partially expanded `Term`s). 

The final `Node` is shipped directly to babel (via `transform.fromAST`) because ES6. All sweet.js code you write is now ES6 (or at least as much as babel can support).

## Recursive descent enforest

The old `enforest` was weird and complicated and basically a giant `if` block. The new `enforest` is still weird and complicated but at least it's a bit more modular. 

## Immutable.js

Currently the expansion algorithm is written as if we were using lists when in fact it's arrays all the way down. Lot's of calls to `concat` that should not be happening if we care about performance.

Now we're using immutable.js lists of syntax objects which should be better.

## New syntax transformer type

There's a new syntax declaration form (analogous to `var`/`let`/`const` for compiletime values) that looks like `syntax <id> = <expr>`. Previously sweet supported a couple different primitive `macro` forms but you could only really put macros into the compiletime environment. This new form allows you to put whatever you want into the env.

Normally (e.g. Racket, old sweet) syntax transformers (aka macros) are just functions but I'm changing things up. Macros are actually objects with two methods `match` and `transform`.

```js
syntax m = {
    // List[Syntax] -> {subst: Substitution, 
    //                  rest: List[Syntax] }
    match: function(stxl) {
        return {
            subst: [],
            rest: stxl.rest()
        };
    },
    // Substitution -> List[Syntax]
    transform: function(subst) {
        return syntaxQuote { 42 };
    }
};
```

The reason for breaking matching and transforming out into two functions is hygiene. Currently sweet has to pass some hygiene information to primitive macros so that they can mark syntax that they match and their result syntax. This is gross and dangerous; badly behaved primitive macros can mess us hygiene in various ways. By splitting macros into two functions we can pull the hygiene manipulation code back into the expander.

More details to work out here but I think this is the right factoring.

This of course is just the primitive form, declarative `rule` can `case` macro forms will be built on top of this.

## No multi-tokens

Right now you can do things like `macro (number?) { ... }` to create multi-token macros. This massively complicates the expander for not much gain. Hacking the lexical structure of your language can be done with readtables (thanks @jlongster!) so let's do that instead.

## Limit infix macros

Infix macros are cool but maybe too cool. The enforestation of operators is massively complicated because we want to allow infix macros to be very flexible. Some heroic work was done by @natefaubion here but I think we are both of the opinion it's not actually worth it. 

I think we can still have them but just in a restricted state. My proposal is that they can only match on previously seen `Term`s and operators create implicit delimiters that infix macros can't "see" out of (so in `2 + inf 42` the `inf` macro sees an empty prefix list). Just my initial intuition, more details to work though.


natefaubion:
I think overall, this is a much needed change, and things are looking a lot better from what I've glanced at.

> Right now you can do things like macro (number?) { ... } to create multi-token macros. This massively complicates the expander for not much gain. Hacking the lexical structure of your language can be done with readtables (thanks @jlongster!) so let's do that instead.

I don't agree with this. Without multi-token macros (or some equivalent) it makes it impossible to create (importable) custom symbolic operators. There may be a better of doing it than before, but I think if we are going to keep custom operators, we need a way to keep this around. I don't think read tables are a good solution for this. Read tables are neither modular nor composable. If we want read tables to be used, then I would like to see them used more along the lines of Racket's `#lang` declaration with an expanded API for controlling the expansion of a module.

> I think we can still have them but just in a restricted state. My proposal is that they can only match on previously seen Terms and operators create implicit delimiters that infix macros can't "see" out of (so in 2 + inf 42 the inf macro sees an empty prefix list). Just my initial intuition, more details to work though.

I strongly agree with this. If you make it a single term look behind only (enough to implement arrows), you don't need to keep the nasty zipper-like structure we were using at all.

> Macros are actually objects with two methods match and transform.

I don't really like the ad-hoc record. I think we could still keep the single function definition if we had a safe API for extracting syntax from the context rather than manipulating it directly, but I don't know what that would look like. In general, I think it would be good to think about the low-level APIs we need to build things like declarative macros as libraries rather than hard wiring them into our own opinionated system. I think it would be great if sweet.js was just a pure low-level core, but that's just me.

If we did really want to go this route, I think it would be better to go all-in with ES6 classes, and have something like a `Macro` base class that you can import from some magic-module, where appropriate hooks are exposed. This also makes it easier to pickup which compile-time values are actually macros. Maybe you could also use this to expose AST-transformers rather than just pure syntax transformers. I don't know what that would look like though, just a thought.

disnet:
>  Read tables are neither modular nor composable. If we want read tables to be used, then I would like to see them used more along the lines of Racket's #lang declaration with an expanded API for controlling the expansion of a module.

Good points. I'm adding a todo item to spec out what we really need for this. I'd really like to keep multi-token code out of the main expander so my hope is that modules are sufficient. We'll see.

>  If you make it a single term look behind only (enough to implement arrows), you don't need to keep the nasty zipper-like structure we were using at all.

Is there any reason to cap it at a single term? 

> I think it would be great if sweet.js was just a pure low-level core, but that's just me.

Totally agree. My thinking is that syntax transformers should be as low-level as possible but not lower than hygiene. Hygiene should be completely handled by the expander, which means that the expander needs to (at minimum) distinguish between syntax provided to the transformer and syntax returned by the transformer.

You're right though, we could still accomplish this with just a single function if the syntax extraction API is "safe". 

Hmm...actually the `match`/`transform` approach has another problem which is that the `rest` syntax returned from `match` is not guaranteed to be correlated with the syntax it put into the substitution (e.g. match could grab all the syntax but return exactly the list provided to it). A buggy `match` could do some damage.

Ok, so maybe like you said we just need a good and safe extraction API, which would obviate the need to break things out into a pair of functions.

Here's some first thoughts. 

Have macros take as their only argument an iterator-like object (so macros don't get direct access to the syntax list, just the iterator). Each time `next` is called the returned value can be appropriately marked by the hygiene algorithm and when the macro returns the expander knows exactly where the "rest" of the syntax is (whatever was not consumed by the iterator).

So the type of a transformer is something like `Iterator -> [Syntax]`.

```js
syntax id = function(ctx) {
  let macro_name = ctx.next();
  let arg = ctx.next();
  return [arg];
}
```

A critical part of the API of course is matching against bits of the grammar (like `:expr`) so the iterator could also have methods like `nextExpression` that call back into `enforest` to build up an expression.

```js
// { mylet $id:ident = $init:expr }
// =>
// { var $id = $init }
syntax mylet = function(ctx) {
  let result = [makeKeyword("var")];
  let macro_name = ctx.next();
  
  let id = ctx.nextIdentifier();
  result.push(id);
  
  let eq = ctx.nextPunctuator();
  result.push(eq);
  
  let init = ctx.nextExpression();
  result.push(init);
  return result;
}
```

This raises the question what should the type of `next*` be. Currently when sweet sees `:expr` it builds up a term via enforest and the destructs it down to an array of syntax and then wraps that syntax in a `()` to keep precedence right (this is confusing to macro authors and leads to bugs). It would probably be better if we could skip the destruct and just have the `next*` functions have the type `() -> Term`. 

Doing this means that a syntax transformer would now have the type `Iterator -> [Syntax or Term]` and so the expander would need to handle both syntax objects and terms in the result of a macro.

Does this seem reasonable?

zackp30:
Damn I'm looking forward to this. :smile:

jlongster:
I saw the refactor branch before but somehow missed this PR. This is huge! My favorite part is no more destructuring and building on top of babel. I always hated how much unnecessary CPU time we were using up with all that.

I don't have strong opinions about the new feature-set, as I'm currently not as involved, but man I can't wait to start playing with this again.

m1sta:
Very excited. In the emerging re-design will the ability to create whitespace-sensitive macros be more accessible?

disnet:
Not directly part of the redesign but it will definitely set us up in a good position to start working on that.

disnet:
So exciting news! I think we're really close to an MVP for the redesign.

I've published an alpha-tagged version on npm so if you would like to try it out and help find bugs etc. get it at `npm install sweet.js@1.0.0-alpha.0`.

This is definitely a MVP so not all the equivalent features from the current `0.7.x` version are here yet but I think we now have a solid base to iterate on. My current plan is to shake out the bugs and then get it merged into master. Then we can start working on the features from the deferred list.

I started on the documentation and you can read it [here](https://rawgit.com/mozilla/sweet.js/redesign/doc/book/_book/index.html).

This example illustrates the highlights:

```js
// create a syntax transformer (aka macro) with the new `syntax` keyword, which
// is like `let` but for compile-time values
syntax function = function(ctx) {
  // ctx is an iterable that provides access to syntax at the call-site
  let id = ctx.next().value;
  let parens = ctx.next().value;
  let body = ctx.next().value;

  // `of` creates a new context around a delimiter syntax object like parens
  let parenCtx = ctx.of(parens);
  let paren_id = parenCtx.next().value;
  parenCtx.next() // =
  // calling `next` with a string invokes the enforester to match
  // against the grammar
  let paren_init = parenCtx.next('expr').value;

  let bodyCtx = ctx.of(body);
  let b = [];
  // contexts are iterables so we can use for-of!
  for (let s of bodyCtx) {
    b.push(s);
  }

  // syntax templates via #``
  return #`function ${id} (${paren_id}) {
    ${paren_id} = ${paren_id} || ${paren_init};
    ${b}
  }`;
}

function foo (x = 42) {
  console.log(x);
}
```

Deferred:

- Pattern matching (aka macro rules)
- Infix
- Custom operators
- Macroclass
- Invoke
- Local expand
- Syntax function zoo
- Source maps
- Macro stepper
- Read tables
- Multi token macro names
- Full Modules (import for macros)

vendethiel:
Wow! That's an insane body of work. Well done!

zackp30:
Yay! I was just thinking about this today when fiddling with JS! :D

acterhd:
When you planned complete the work?

disnet:
@vendethiel @zackp30 thanks for the props!

@acterhd software estimates are fiction and doubly so for open source so I won't bother telling you a tall tale. It goes faster when people contribute though so feel free to submit pull requests :)

disnet:
@vendethiel :)

disnet:
So in super-cool news the redesign is now self-hosting! Sweet.js is once again building sweet.js!

This means that I'm calling the redesign officially close enough for merging into master. I'm planning on waiting a week to see if any blocking bugs show up but after that I think we can start iterating on a much better foundation.

The pre-release version is available under the `pre` npm tag (i.e. install it with `npm install sweet.js@pre.0`). The documentation is available [here](https://rawgit.com/mozilla/sweet.js/redesign/doc/book/_book/index.html).

Alexsey:
Link from previous post '<a href='https://rawgit.com/mozilla/sweet.js/redesign/doc/book/_book/index.html'>here</a>' doesn't work :(

disnet:
@Alexsey I split the documentation out into the [tutorial](https://github.com/mozilla/sweet.js/blob/redesign/doc/1.0/tutorial.adoc) and the [reference](https://github.com/mozilla/sweet.js/blob/redesign/doc/1.0/reference.adoc).

vendethiel:
\o/

Issue #233, opened 2014-02-01
Integrate sweet.js with ES6 modules


disnet:
There are two papers on macros and modules that are probably relevant. [Composable and Compilable Macros](http://www.cs.utah.edu/plt/publications/macromod.pdf) and [Implicit Phasing for R6RS Libraries](http://www.cs.indiana.edu/~dyb/pubs/implicit-phasing.pdf).

The basic idea of both papers is that you have to track the phase level any bit of syntax currently is in. The phase isn't as simple as compile-time vs. run-time because a case macro (compile-time) might need to import and use another macro (earlier compile-time) which might also need to use a macro (even _earlier_ compile-time) and so on. 

``` js
// bar.js
export macro bar {
    case {_ ()} => {
        // ...
    }
}
```

``` js
// foo.js
import { bar } from "bar"
export macro foo {
    case {_ () } => { 
        bar()
        // ... 
    }
}
```

``` js
// main.js
import { foo } from "foo"
foo()
```

The papers outline two approaches we could take. The first, following the "Composable and Compilable" paper and Racket, involves explicitly declaring what phase level an import is for. 

``` js
import { $ } from "jquery"
import forSyntax { _ } from "underscore"

let m = macro {
    case { } => {
        $("#foo") // error: $ not bound
        _.map([1,2,3], function (x) { ... });
        // ...
    }
}
$("#foo") // works fine
_.map(...) // error: _ is not bound
```

This gets ugly though when you have a macro definition in your macro definition since you need to import up two phase levels:

``` js
import { $ } from "jquery"
import forSyntax { _ } from "underscore"
import forSyntax forSyntax { ld } from "lodash"
// or maybe following racket:
import meta 2 { ld } from "lodash"

let m = macro {
    case { } => {
        let n = macro {
            case {} => {
                ld.map(...)
                // ...
            }
        }
        $("#foo") // error: $ not bound
        _.map([1,2,3], function (x) { ... });
        // ...
    }
}
$("#foo") // works fine
_.map(...) // error: _ is not bound
```

The "Implicit Phasing" paper gives a technique for tracking all the phasing implicitly so you wouldn't need to do `import forSyntax`. The paper seems to suggest that there are no real trade offs in doing this but I seem to remember @dherman mention there was a good reason Racket stayed with explicitly phase labeling (something about the REPL maybe?). 

I would prefer to keep the phasing implicit if possible since forcing macro authors to think about the details of module phasing seems not great.

[...]

PR #518, opened 2016-03-28
Add MacroContext::read

bopjesvla:
Takes a syntax type or grammar production, a value and an error callback. All arguments are optional, although the type is necessary to check the value. Returns the next syntax object.

Passes a SyntaxError to the callback if the type or value doesn't match, a TypeError if the syntax type doesn't exist, and a ReferenceError* if the end of the context is reached. If no callback is given, errors are thrown.

\* a SyntaxError might make more sense, but this keeps error handling simple

-----

This could also be implemented as an extension of the "next" method.


disnet:
Ah cool, was thinking we needed something like this :)

Few thoughts.

First, the name `read` won't work because that's [overloaded](https://github.com/mozilla/sweet.js/blob/master/src/shift-reader.js) in Sweet. Maybe `match`?

Second, I don't like passing strings for the is* methods. Much prefer being explicit. Maybe a method per syntax type? This avoids needing to throw a `TypeError`.

```js
matchIdentifier(value?)
matchStringLiteral(value?)
//...
```

Also, don't commit the built files in `dist/`. We'll make a build after a PR lands on master. And, you'll probably want to add tests [here](https://github.com/mozilla/sweet.js/blob/master/test/test_macro_expansion.js).

bopjesvla:
> First, the name read won't work because that's overloaded in Sweet. Maybe match?

Mhm, extending `next` makes the most sense to me. My current `read` implementation behaves the same for 'expr' and no arguments, except for the fact that reading when done results in an error.

> Second, I don't like passing strings for the is* methods. Much prefer being explicit.

I should've addressed this before creating this pull request. I'm not a fan of the is* methods, to be honest. I'd prefer one `isKind` method taking a string and an optional value. It's more dynamic, which is key for something that's supposed to be built on.

Right now, if someone wants to create a simple custom read function like mine, they either have to produce a function for every type like you're suggesting or use the "is"+type trick I used.

The same thing goes for creating an abstract syntax-producing syntax, I think. In a `macro`-like syntax, directly passing the type in `$ident:type` to isType as a string would be most preferable.

Also, at some point you might want to allow people to add their own syntax types. A dynamic approach would only require Types[typeName] = {matcher: fn} to be set. The approach you're suggesting would require adding "is"+typeName and "match"+typeName to the MacroContext prototype.

In the dynamic approach, adding MacroContext::typeExists(type) would be as easy as checking if Types[typeName] exists.

> Maybe a method per syntax type? This avoids needing to throw a TypeError.

I find `TypeError: Brackkets is an invalid type` more descriptive than `TypeError: ctx.isBrackkets is not a function` so that shouldn't be a deal breaker.

> Also, don't commit the built files in dist/. We'll make a build after a PR lands on master. And, you'll probably want to add tests here.

Will do. I'll also remove the babel-fix merger.

disnet:
> Mhm, extending next makes the most sense to me. My current read implementation behaves the same for 'expr' and no arguments, except for the fact that reading when done results in an error.

Note quite since it returns the value, not an `IteratorResult`. `next` is part of the iterator API and making it return something other than an `IteratorResult` in certain cases or throwing exceptions is a non-starter imo. I like `match` but would be open to anything besides `read` and `next`.

> I'd prefer one isKind method taking a string and an optional value. It's more dynamic, which is key for something that's supposed to be built on.

Fair point.

I suppose this goes to the question of what is the macro context object? Long term if we can get away with it, I really don't want *any* methods besides `next` available on it. Much better to build helper functions in a separate module that work over a context and then `import ... for syntax` them. However, this requires #233 to be implemented first so should we then hold up important functionality like this?

Probably not, so let's keep it low-level like you've proposed and once #233 lands we can build on top of it.

bopjesvla:
> Note quite since it returns the value, not an IteratorResult. next is part of the iterator API and making it return something other than an IteratorResult in certain cases or throwing exceptions is a non-starter imo. I like match but would be open to anything besides read and next.

Oh man, I didn't realise this fully corresponds to ES6 standards. How about this?

-----

`next` behaves as it does right now when it's presented with zero args or a grammar production.
But when it's given a type and an optional value, it's a shorthand for

```
let next = ctx.next();
if (next.done || !next.value.isKind(type, value))
  throw SyntaxError("descriptive message");
return next;
```

This still throws, but only when it's clearly meant to. I agree that having nothing but the `next` method would be nice, yet I feel like the "read this or die" pattern is common enough to be in the core.

disnet:
Yep, worth having in core but let's not call it `next` :)

bopjesvla:
After some contemplation I think that until #233 is implemented, we should give this a rest. This hack can be used for now:

```
syntax addReader = ctx => {
	let MacroContextPrototype = Object.getPrototypeOf(ctx)
	
	MacroContextPrototype.read = function(type, value, errorHandler) {
		...
	}
	
	return #``;
}

addReader
```

disnet:
Cool, #531 should land before too long and we can do all this in libraries. Thanks!

PR #526, opened 2016-04-01
Allow macro names to be punctuators

disnet:
Allow macro names to be punctuators. Also allows tokenization of `@` (this was legal pre-1.0). Allows you to do fun things like:

```js
syntax @ = function (ctx) {
  return #`this`;
}
@
```

### DIFF
--- a/src/enforester.js
+++ b/src/enforester.js
@@ -765,10 +765,10 @@ export class Enforester {
     });
   }
 
-  enforestBindingTarget() {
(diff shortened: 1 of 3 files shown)


tycho01:
Hah. I got `#` to work before, only now realized EcmaScript doesn't consider it a punctuator for some reason. It's hard to Ctrl-F the docs to figure out how it is classified though. :)

PR #530, opened 2016-04-12
Renaming is* and from*, refactoring

bopjesvla:
Changed `isSomeType(value)` to `match("someType", value)` on Syntax and SyntaxOrTermWrapper, `isKind("someType", value)` on Enforester. They can all take a regex as a value (untested for now).

`Syntax.fromSomeType(val, stx)` has become `Syntax.from("someType", val, stx)`.


disnet:
I strongly prefer explicit methods rather than passing strings.

The `is*` methods on `Syntax` could be refactored to include checking the value though. That has been bugging me for a while.

bopjesvla:
Explicit methods really don't make sense to me, both from a macro writer's and a developer's perspective. As I said in #518, passing a string is more dynamic, which is key for something that's supposed to be built on.

As it stands, dynamically matching a syntax type requires including `"is"+capitalize(type)` somewhere, which is just plain weird.

Having all syntax types live in a single object also has clear advantages for development and extendability. In the refactored code, all syntax types just consist of a `match` and a `create` function (although `create` functions of some subtypes like literal booleans are still missing). All other match-like functions are passed to the Syntax object. Also, `Syntax.from(type, value).match(type, value)` is now always true for valid types.

What is it that you like about explicit methods?

disnet:
Let me rephrase my objection :)

I strongly prefer *using* explicit methods (especially in the internals of Sweet) rather than having *only* the dynamic `match` method. The `is*` methods can obviously be defined in terms of `match`, but I still want them to exist and to be used in, for example, the enforester.

bopjesvla:
Well, since the the whole point of refactoring was to make matching play better with macros, we can just add an internal macro:

```
syntax is = ctx => {
	let type = ctx.next().value,
		args = ctx.next().value.inner(),
		stx = args.next("expr").value,
		seperator = args.next().value,
		value = args.next("expr").value,
		Syntax = #`;`.get(0).constructor,
		typeString = Syntax.from("string", type.val())
		
	if (value) {
		return #`${stx}.match(${typeString}, ${value})`
	}
	else {
		return #`${stx}.match(${typeString})`
	}
}

if (is braces(stx)) ...
if (is number(stx, 10)) ...
```

When custom operators are back, we can turn this into `stx is number(10)`.

disnet:
Why use a macro when a function works just as well?

bopjesvla:
A single macro is less bloated than a function for every type. Also, it's dynamic. If you add a type to syntax.js, you won't have to add any methods to the enforester.

The explicit methods don't add any functionality. They're just syntactic sugar. That's what macros are meant for.

vendethiel:
Macro would allow for compile-time checking of the argument (so technically, it's even better than `.isXYZ()` methods that can be typo'd :P ).

disnet:
Obviously I like macros 😄 but their complexity doesn't pay their way in this case.

bopjesvla:
Heh, I expected you to jump on this because of the macro. Do you understand why I dislike the current implementation?

Imagine that instead of `document.createElement`, the folks at Ecma gave us `document.createDiv`, `document.createSpan`, etc. This works fine for people who just need to create divs and spans, but it's a pain in the ass for people who want to write libraries like jQuery and React on top of JS, because they have to route every single tag name, instead of just passing an argument through to `document.createElement`.

-----

As for the enforester, currently `is*` methods live on the prototype, like this:

```
class Enforester {
  ...
  isNumericLiteral(term) {
    return term && (term instanceof Syntax) && term.isNumericLiteral();
  }

  isStringLiteral(term) {
    return term && (term instanceof Syntax) && term.isStringLiteral();
  }

  isTemplate(term) {
    return term && (term instanceof Syntax) && term.isTemplate();
  }
  ...
}
```

This makes little sense considering that enforester instances have nothing to do with syntax matching.

-----

As for a macro not being suitable for this situation, I frankly find it hard to believe that's the position you hold as the maintainer of sweet.js, and I suspect it's mostly familiarity bias speaking. If you don't want to use a small macro for eliminating something as repetitive and verbose as the above code, then what _should_ we use macros for?

disnet:
> Imagine that instead of document.createElement, the folks at Ecma gave us document.createDiv, document.createSpan, etc.

Bad analogy. I’m not saying don’t have `isKind`/`createElement`. I’m saying have both `isKind` and the convenience methods `is*` (`createDiv`, `createSpan` etc.).

> As for the enforester, currently `is*` methods live on the prototype

Yes that is wonky. A broader refactoring is certainly warranted.  

> If you don't want to use a small macro for eliminating something as repetitive and verbose as the above code, then what should we use macros for?

There is long-standing advice in the Scheme world to use macros only when functions don’t suffice. To my mind it only makes sense to reach for the power of macros if you want to introduce new syntax (something that can’t be expressed via a function) or to relieve some “onerous” repetition. What counts as “onerous” is admittedly a judgment call and I have judged the token types (all 15 of them) and found them wanting 😄.

To me seeing:

	isIdentifier(value) {
		return this.match('identifier', value);
	}
	...

in the codebase is much more readable than trying to understand the macro that generates the equivalent.

bopjesvla:
> Bad analogy. I’m not saying don’t have isKind/createElement. I’m saying have both isKind and the convenience methods is* (createDiv, createSpan etc.).

I can live with that. I took your reaction as "yeah, it's possible but not necessary".

But still:

> To my mind it only makes sense to reach for the power of macros if you want to introduce new syntax

The plan is to introduce a new operator, although that's not yet possible.

> all 15 of them

That's 45 functions across 3 modules.

> in the codebase is much more readable than trying to understand the macro that generates the equivalent.

When pattern-based macros land, this should no longer be the case. For now, we can just add an example in the comments.

bopjesvla:
Simple mapping of the syntax types, no macros involved.

disnet:
So I think the key utility you identified is the `isKind("somekind", value)` method on `Syntax` and `SyntaxOrTermWrapper`. That is definitely valuable for code that needs to dynamically select the kind of syntax it is matching. If you want to update this PR to just add that method on `Syntax`/`SyntaxOrTermWrapper` (while leaving the existing `is*` methods intact) I would be willing to accept it. Let’s defer changes to `Enforester` to a separate issue/PR discussion.

disnet:
Oh cool! Didn't see you updated this PR. You need to add a comment when changes are made otherwise I don't get notified :)

I'll try to review soon.

src/macro-context.js
```diff
@@ -52,48 +34,75 @@ export class SyntaxOrTermWrapper {
     this.context = context;
   }
 
-  isKeyword(value) {
-    return isKeyword(this[symWrap], value);
+  match(type, value) {
+    let stx = this[symWrap]
```
disnet: Our style is `;` terminated lines.

src/syntax.js
```diff
@@ -18,6 +18,156 @@ function sizeDecending(a, b) {
   }
 }
 
+export let Types = {
+  // types
+  null: {
+		match: token => !Types.delimiter.match(token) && token.type === TokenType.NULL,
```
disnet: odd spacing

src/syntax.js
```diff
@@ -18,6 +18,156 @@ function sizeDecending(a, b) {
   }
 }
 
+export let Types = {
+  // types
```
disnet: remove redundant comment

disnet:
Can you rebase and squash?

bopjesvla:
You should be able to squash this PR from the web UI, see https://help.github.com/articles/configuring-pull-request-merge-squashing/.

Issue #223, opened 2014-01-28
Failed to introduce macro with ellipsis from inside other macro


andreypopp:
I see the following error

```
SyntaxError: [patterns] The pattern variable is not bound for the template
12:                 rule { { $stmt ... } } => {
```

when trying to declare macro

```
macro outer {
    case { $name } => {
        return #{
            macro ok { // this works!
                rule { $id } => { $id(true) }
            }

            macro fail { // this doesn't work!
                rule { { $stmt ... } } => {
                    function x() {
                        $stmt ...
                    }
                }
            }
        }
    }
}

outer
```

PR #531, opened 2016-04-13
Modules and Phasing

disnet:
Issue #223.

This is not ready to be merged just yet but I’ve been making some good progress on getting proper phasing and module support working and wanted to give people a heads up on what is coming.

The approach I’m taking is heavily modeled after how Racket does it as described in [Composable and Compilable Macros](https://www.cs.utah.edu/plt/publications/macromod.pdf). This means:

## Explicit Phasing

Import statements declare which phase they want to import into. So:

	import { m } from 'm';  // bound for runtime
	import { n } from 'n' for syntax;  // bound for compiletime
	m;
	n; // unbound
	syntax s = function (ctx) {
		m; // unbound
		n;
		// ...
	}

## Multiple Instantiation

There is a tower of phases (a module that is `import ... for syntax` might have a `import ... for syntax` which might have a `import ... for syntax` which might have…) and a module is separately instantiated in each phase. This is implemented via node’s [vm.runInContext](https://nodejs.org/api/vm.html#vm_vm_runincontext_code_contextifiedsandbox_options).

## Common Concerns

Unfortunately there are at least four different kinds of modules we need to support:

- CommonJs modules (`require` and `exports`)
- node compiled modules (`.node`)
- ES2015 modules
- Sweet modules (ES2015 with syntax definitions)

Obviously Sweet modules are the easiest to handle, just `compile` and then `invoke` or `visit` as appropriate. We can recognize Sweet modules with the `#lang` pragma and then use the standard pipeline. 

Dealing with the other kinds of modules is more tricky. 

ES2015 modules are conceptually straightforward (just `compile` like normal) but *knowing* that a file is ES2015 is complicated right now because the community hasn't settled on a standard way to do that. There is an ongoing [node discussion](https://github.com/nodejs/node/wiki/ES6-Module-Detection-in-Node) where it sounds like they are converging on a file extension (*sigh* which is wrong, I think rollup got it right by using `package.json`, but whatever). Anyway, it's not settled yet. 

The important one is commonjs since at the moment almost everything is commonjs. I think it is straightforward to just plop a commonjs module inside a `vm.runInContext` with hooks into the exports object but I haven’t tried yet.

I’m pretty sure we can’t do anything about `.node` compiled modules besides `require`ing them. This is probably fine?

/cc @elibarzilay @natefaubion


elibarzilay:
@disnet: One comment I have about this is that it might be good to consider some syntax that would be more uniform.  For example, something like

    for_syntax import { n } from "n";

My point here is that it would be nice to avoid a situation where there *is* later some way to run code at the syntax level (eg, some `begin_for_syntax`), and then working hard to make sure that

    begin_for_syntax import { n } from "n";
    import { n } from "n" for syntax;

are both the same.

natefaubion:
> My point here is that it would be nice to avoid a situation where there is later some way to run code at the syntax level (eg, some begin_for_syntax), and then working hard to make sure that

>    begin_for_syntax import { n } from "n";
>    import { n } from "n" for syntax;

> are both the same.

I don't think this makes sense. `import`s should always be at the top level of a module, not scoped under any hypothetical `begin_for_syntax`.

:+1: On the proposed syntax in the PR

I don't have much to add wrt to importing ES modules vs CommonJS. I think in old discussions we landed on `package.json` as well, but if that's not where the puck is going, then... _shrug_

elibarzilay:
@natefaubion (1) Yeah, Racket started with that assumption too, which is how it ended with the salad it has now.  (2) a `begin_for_syntax` is hypothetical of course (and the `begin` part makes no sense in JS), but the functionality is very much not so (it's essentially the RHS of a syntax definition).

disnet:
So more to be done with modules and phasing but I think this PR is good enough for a checkpoint merge to build on top of. What I have now is not fully general; only modules that have a `#lang` pragma can be imported for syntax.

Left to do are integration with non-sweet ES2015 modules and legacy CommonJS modules.

PR #542, opened 2016-04-22
Move CLI defaults to the API

bopjesvla:
This enables compiling through the API without specifying a loader, resolver, transformer, etc.


bopjesvla:
Travis is using the new CLI tool with the old dist. This can be fixed by pushing the build or by adding the CLI tool to the build process, your call.

disnet:
The reason `src/sweet.js` does not specify a default loader/resolver is so we can use `src/sweet.js` in both the browser (`browser/scripts/sweet.js` is built from `src/sweet.js` via `npm run build:browser`) and the node cli.

If you really want a node specialized version of the API you could make something like `src/sweet-node.js` but I'll probably be reworking most of loading/resolving in #531.

bopjesvla:
This seems easier to fix by taking advantage of the browser field in package.json. I'll type something up in a bit.

disnet:
CI failed on this. Are you still working on this? I'm not convinced the changes you were making would work but I'm not a webpack expert.

bopjesvla:
That's because the updated CLI tool depends on the updated API. There are two ways forward from here:

> This can be fixed by pushing the build or by adding the CLI tool to the build process, your call.

disnet:
Ah, go a head and add a build commit.

bopjesvla:
a comment

src/sweet.js
```diff
@@ -13,22 +13,29 @@ import BindingMap from "./binding-map.js";
 import Term from "./terms";
 import { Modules } from './modules';
 
+// not available in browser
+
+import { transform as babelTransform } from "babel-core"
```
disnet: add `;`

disnet:
There are merge conflict. Could you resolve them?

bopjesvla:
Done!

disnet:
Thanks!

Issue #538, opened 2016-04-21
Add .name to macro context


disnet:
Need a way of retrieving the invoking macro's name syntax object:

```js
syntax m = ctx => {
  ctx.name(); // m
}
m 42;
```

bopjesvla:
I think you mentioned in #529 that unlike `#'m'`, using `ctx.name()` would make `m` recursive (edited for this example):

```
Definitely, that was in my mind but didn’t make it into code for the MVP. Trivial to add but not critical until we have syntax creation functions and the need to perform hygiene shenanigans.

In this case though, you don’t want ctx.name() since that would just make it a recursive definition. You want the previous binding via #'m'.
```

Wouldn't this make `syntaxrec` superfluous?

disnet:
`.name` gets the syntax object at the call-site (along with its associated scope) while `syntaxrec` gets the syntax object at the definition site. The call-site syntax object is often used in Scheme/Racket for hygiene bending:

```js
import { makeIdentifier } from 'syntax' for syntax; // or something similar

let x = 'outer';
syntax m = ctx => {
    return #`${makeIdentifier('x', ctx.name()}`;  // makes a new identifier bound in the scope of ctx.name()
}

function foo(x) {
  let x = 'inner';
  m; // expands to the inner x binding
}
```

bopjesvla:
Ah, that makes sense, thanks!

disnet:
Closed with #544

PR #544, opened 2016-05-06
Macro context name via method

gabejohnson:
Exposes the macro context name via a method.
```javascript
syntax foo = ctx => {
  ctx.name().val() // foo
  //...
}
foo 42
```
Implements #538


disnet:
Thanks!

Issue #519, opened 2016-03-28
Nested syntax templates


bopjesvla:
```
syntax helper2 = ctx => #`
	syntax test2 = ctx => {
		return #`1`
	}`

helper2

syntax helper1 = ctx => #`(#`1`)`

syntax test1 = ctx => helper1

test1
```

Neither works.

disnet:
The escape syntax for syntax templates is the same as normal templates:

```js
#`a literal backtick \` and then we're done`
```

disnet:
Oops. That's supposed to be the escape but looks like we have a bug:

```js
syntax test = function (ctx) {
  return #`let x = \`template string\`;`
}
test
```

throws:

```
/Users/tim/dev/sweet.js/node_modules/shift-parser/dist/tokenizer.js:731
            throw this.createILLEGAL();
            ^

Error: [2:21]: Unexpected "`"
    at JsError.Error (native)
    at new JsError (/Users/tim/dev/sweet.js/node_modules/shift-parser/dist/tokenizer.js:161:81)
    at Reader.createError (/Users/tim/dev/sweet.js/node_modules/shift-parser/dist/tokenizer.js:277:14)
    at Reader.createILLEGAL (/Users/tim/dev/sweet.js/node_modules/shift-parser/dist/tokenizer.js:245:53)
    at Reader.getEscapedIdentifier (/Users/tim/dev/sweet.js/node_modules/shift-parser/dist/tokenizer.js:731:24)
    at Reader.scanIdentifier (/Users/tim/dev/sweet.js/node_modules/shift-parser/dist/tokenizer.js:795:63)
    at Reader.advance (/Users/tim/dev/sweet.js/node_modules/shift-parser/dist/tokenizer.js:1437:25)
    at Reader.advance (/Users/tim/dev/sweet.js/build/src/shift-reader.js:325:90)
    at Reader.read (/Users/tim/dev/sweet.js/build/src/shift-reader.js:227:24)
    at Reader.read (/Users/tim/dev/sweet.js/build/src/shift-reader.js:252:28)
```

bopjesvla:
Yeah, I tried that too. I based the syntax for nested templates on the reader tests.

gabejohnson:
This is still an issue. I have some ideas, but they involve changing our strategy for reading syntax templates. I'd also like to implement the ideas from https://github.com/sweet-js/sweet.js/pull/551#issuecomment-240721330 at the same time.

PR #551, opened 2016-06-02
Began Work on Fixing Nested Syntax

belph:
**NOTE: This PR is not ready to merge!**

I have begun working on fixing #519, but I have reached a bit of a stumbling block. I have gotten [`shift-reader.js`][shift-reader-master] to accept files with nested syntax, but something is getting garbled in between (details below), leading me to believe that I'm missing some step in the expander pipeline.

To explain where I'm at, here's what I know:

- The reading of syntax templates (and, well, everything else) is done in `src/shift-reader.js`.
- Syntax objects begin when `src/shift-reader.js`'s `advance` function ([master][advance-master]) returns a token of type `LSYNTAX`.
- The contents of syntax objects are read like normal syntax; the only thing that makes them special
is that the handler for `LSYNTAX` bundles all of the read pieces together into an immutable list.
- The reason for the error that is mentioned in the bug report is because `advance` in `src/shift-reader.js` does not handle the ``"`"`` character (or it's escaped variant), so it delegates to the superclass (which also does not handle it). 

Thus, I figured that fixing the error would just be a matter of adding the proper escapes to `advance`, allowing nested syntax objects to be read correctly. As far as reading the syntax goes, this all seems to work correctly. *Note: This change is what the first commit of this PR reflects.*

So, what happened? I tried to run this modified version on the following code:
```
syntax makeSayer = function(ctx) {
  let x = ctx.next().value;
  return #`syntax ${x} = function(ctx) { return \`console.log('hello world')\`; }`;
}

makeSayer hi
```

The result was the following error messsage:
```
/home/belph/git-belph/sweet.js/dist/syntax.js:174
        return s_701.addScope(scope_696, bindings_697, options_698);
                    ^

TypeError: Cannot read property 'addScope' of null
    at /home/belph/git-belph/sweet.js/dist/syntax.js:174:21
    at /home/belph/git-belph/sweet.js/node_modules/immutable/dist/immutable.js:3018:46
    at List.__iterate (/home/belph/git-belph/sweet.js/node_modules/immutable/dist/immutable.js:2208:13)
    at IndexedIterable.mappedSequence.__iterateUncached (/home/belph/git-belph/sweet.js/node_modules/immutable/dist/immutable.js:3017:23)
    at seqIterate (/home/belph/git-belph/sweet.js/node_modules/immutable/dist/immutable.js:606:16)
    at IndexedIterable.IndexedSeq.__iterate (/home/belph/git-belph/sweet.js/node_modules/immutable/dist/immutable.js:322:14)
    at IndexedIterable.mixin.toArray (/home/belph/git-belph/sweet.js/node_modules/immutable/dist/immutable.js:4260:23)
    at new List (/home/belph/git-belph/sweet.js/node_modules/immutable/dist/immutable.js:2067:62)
    at reify (/home/belph/git-belph/sweet.js/node_modules/immutable/dist/immutable.js:3572:37)
    at List.mixin.map (/home/belph/git-belph/sweet.js/node_modules/immutable/dist/immutable.js:4403:14)
```

Let's first dissect the syntax object in question as follows:

```
<ret> = #`syntax ${x} = function(ctx) { \`console.log('hello, world')\`; }`
                                        |------- <ret_nested> --------| 
```

Now, here's what I can tell happens:

- `read` ([my fork's copy][read-nested-syntax]) will tokenize `<ret>` and `<ret_nested>` just fine; `<ret>` is an immutable list with an entry at index-whatever that is another immutable list corresponding to <ret_nested>
- Sometime between then and the return value of the call to `deserializer.read(str)` in `loadForCompileTime` ([my fork's copy][deserializer-nested-syntax]), the location of `<ret_nested>` in `<ret>` is set to `null`. Why? Not a clue.

And that is where I'm at. Am I missing some step here? I feel like this is, if nothing else a step in the right direction; nested syntax is a powerful feature to have, so I'd love to get this fixed!

[shift-reader-master]: https://github.com/sweet-js/sweet.js/blob/d236e8fdad4ebfbf69267bb6f24f9cbf21c0ff54/src/shift-reader.js
[advance-master]: https://github.com/sweet-js/sweet.js/blob/d236e8fdad4ebfbf69267bb6f24f9cbf21c0ff54/src/shift-reader.js#L346-424
[read-nested-syntax]: https://github.com/belph/sweet.js/blob/0496c6580c98c531944d57d86a381f09e955193f/src/shift-reader.js#L305-363
[deserializer-nested-syntax]: https://github.com/belph/sweet.js/blob/0496c6580c98c531944d57d86a381f09e955193f/src/load-syntax.js#L47


src/shift-reader.js
```diff
@@ -58,6 +58,9 @@ const binaryOps = ["+", "-", "*", "/", "%","<<", ">>", ">>>", "&", "|", "^",
 
 const unaryOps = ["++", "--", "~", "!", "delete", "void", "typeof", "yield", "throw", "new"];
 
+// Any -> Boolean
+const isNotFalse = R.compose(R.not, R.curry(R.equals)(false));
```
vendethiel:
```
const isNotFalse = R.compose(R.not, R.curry(R.equals)(false));
// vs
const isNotFalse = function (b) { return b != false; }
```

belph: Fair enough :)
I just was trying to emulate the existing style as much as I could. I can change that to a simple function instead.

src/shift-reader.js
```diff
@@ -311,7 +330,7 @@ export default class Reader extends Tokenizer {
 
       if (isLeftDelimiter(tok)) {
         if (isLeftSyntax(tok)) {
-          this.insideSyntaxTemplate.push(true);
+          this.syntaxNestingDepth++;
```
disnet: This actually won't work. The nested arrays of booleans are needed to track syntax templates inside of interpolations:

```js
#`foo ${42 + #`blah`}`  // nonsense but lexically fine
```

belph: Okay, so should a stack with the nesting depth pushed onto it work, then?

disnet:
So I just had a crazy idea. Syntax template are already syntactically special so what if we just don't escape them? Beginning (`` #` ``) and ending (`` ` ``) delimiters are always distinguished so there is no syntactic reason to escape them. Just start a recursive read when you get to a `` #` `` inside a read of an outer syntax template.

Normal templates still need to be escaped of course.

```js
syntax m = ctx => {
  return #`
    syntax n = ctx => {
      let s = \`foo\`;
      return #`
        syntax o = ctx => {
          return #`1`
        }
      `
    }
  `
}
```

This avoids the need to count backslashes and all that.

This approach confuses existing syntax highlighters a little but since all the backticks must be balanced the confusion is localized to the syntax templates themselves.

Does this seem reasonable or am I missing something?

belph:
@disnet That seems perfectly reasonable (it's the same approach Racket takes!)

disnet:
Oh! The trick is nested interpolations:

```js
let s = #`
  let ss = #`foo \${42}`
`;
```

You'll need to count escapes for those I think.

belph:
Okay, so a double escape look like `\\${42}`, but what would a triple escape look like? `\\\${42}` or `\\\\${42}`? If it's the latter, what about using something like Racket's `unquote-syntax` shorthand (`#,`) to avoid an exponential blowup of escapes?

disnet:
How would `#,` help?

belph:
Using the standard escaping for `\`, one can see that `#,` quickly becomes considerably less typing (and is much more descriptive of the unescaping depth)

Depth | Escape sequence with `\` | Escape Sequence with `#,`
------|------|----
0 | `${42}` | `${42}`
1 | `\${42}` | `#,${42}`
2 | `\\\${42}` | `#,#,${42}`
3 | `\\\\\\\${42}` | `#,#,#,${42}`
4 | `\\\\\\\\\\\\\\\${42}` | `#,#,#,#,${42}`

disnet:
Ah! I see what you meant. Yeah do that 😄

gabejohnson:
@disnet if we took care of syntax templates with a combination of reader macros and syntactic extensions, we could do this:
```js
syntax m = ctx => {
  return #`
    syntax n = ctx => {
      let s = \`foo\`;
      return ${ #`
        syntax o = ctx => {
          return ${ #`1` }
        }
      ` }
    }
  `
}
```
This way a `syntaxTemplate` macro could take care of converting nested term objects back into escaped syntax template strings and would be transformed to:
```js
syntax m = ctx => {
  return syntaxTemplate`
    syntax n = ctx => {
      let s = \`foo\`;
      return ${ syntaxTemplate`
        syntax o = ctx => {
          return ${ syntaxTemplate`1` }
        }
      ` }
    }
  `
}
```
Thus `m` gets transformed to:
```js
syntax n = ctx => {
  let s = \`foo\`;
  return ${ #`
    syntax o = ctx => {
      return ${ #`1` }
    }
  ` }
}

```

disnet:
@gabejohnson humm...I'm not following. Could you expand (ha!) on what you meant?

gabejohnson:
My point is that if you just require interpolating if you're nesting syntax templates you can avoid any special syntax and just parallel the syntax used by template literals already:
```js
syntax m = ctx => {
  return #`
    syntax n = ctx => {
      let s = \`foo\`;
      return ${ #`
        syntax o = ctx => {
          return ${ #`${ ctx.next().value }` }
        }
      ` }
    }
  `
}
```
vs.
```js
syntax m = ctx => {
  return #`
    syntax n = ctx => {
      let s = \`foo\`;
      return #\`
        syntax o = ctx => {
          return #\`\\${ ctx.next().value }\`
        }
      \`
    }
  `
}
```
or
```js
syntax m = ctx => {
  return #`
    syntax n = ctx => {
      let s = \`foo\`;
      return #`
        syntax o = ctx => {
          return #`\\${ ctx.next().value }`
        }
      `
    }
  `
}
```
While there are more characters at shallow nesting depths, you don't have to count slashes as you nest.

jlongster:
@gabejohnson I like that, I think it feels the most natural.

gabejohnson:
Something I just thought of. Which context does the innermost `ctx` refer to?

Issue #537, opened 2016-04-21
Add reset to macro context


disnet:
As discussed in #529 we need a way to reset the macro context. A `.reset` is the most obvious:

```js
syntax m = ctx => {
  ctx.next('expr'); // foo + bar
  // oops, just wanted one token
  ctx.reset();
  let {value} = ctx.next(); // foo
  //...
}

m foo + bar
```

Other possible names:
- `.clear`
- `.revert`
- `.rewind`

disnet:
Also worth considering along with this (via @vendethiel) is a non-advancing method like `.peek`.

gabejohnson:
I read through #529. Would it make sense to provide a generator as the argument to the transform function making your example:
```
syntax m = ctxGen => {
  let ctx = ctxGen();
  ctx.next('expr'); // foo + bar
  // oops, just wanted one token
  ctx = ctxGen();
  let {value} = ctx.next(); // foo
  //...
}

m foo + bar
```

I wouldn't think this too difficult as you're using persistent data structures for `stxl` in `Enforester`.

disnet:
Yeah, I thought about that too. Definitely worth considering. Maybe something like:

```js
syntax m = ctx => {
  let name = ctx.name();
  let iter = ctx.iter();
  iter.next('expr');
  let { value } = ctx.iter().next(); // start over

  // once we add infix macros again
  let prev = ctx.prevIter();
  prev.next();
  // ...
}
```

gabejohnson:
Arbitrary lookbehind would be pretty sweet (excuse the pun).

gabejohnson:
Have you thought about using a zipper instead of an iterator for tree traversal?

```javascript
syntax m = ctx => {
  ctx.val(); // m
  ctx.next().val(); // 2
  ctx.prev().val(); // 1
  let outer = ctx.outer();
  outer.prev().prev().val(); // a
  outer.prev().inner().val(); // undefined?
  ctx.val(); // m
}
function a() { return 1 m 2 };
```

I realize that this would be departure and a large implementation change. Just thinking out loud.

gabejohnson:
@disnet, `outer` might be overkill. The use-case I had in mind was Racket-like submodules. An example taken from https://docs.racket-lang.org/guide/Module_Syntax.html#%28part._submodules%29:

```javascript
module land_animals {
  export const lion = "Leo";
  export const tiger = "Tony";
  export const bear = "Yogi";
  export const mouse = "Mickey";
}
[...]

Issue #529, opened 2016-04-11
Extendable syntaxes


bopjesvla:
Extendable syntaxes would allow for:

- the usage of independently written macros with the same name, provided that they're used in different contexts
- the inline extension of an imported macro without forking a module
- anything that requires backtracking, such as the `macro` syntax

The most natural implementation I can think of would be to allow syntax functions to return null, after which the reader tracks back and tries the previous macro with the same name:

```
syntax each = ctx => {
	let variable = ctx.next("expr").value,
		OF = ctx.next().value,
		iterable = ctx.next("expr").value,
		body = ctx.next().value
	
	if (variable && iterable && OF.isIdentifier("of") && body.isBraces()) {
		return #`for (${variable} ${OF} ${iterable}) ${body}`
	}
	
}
syntax each = ctx => {
	let iterable = ctx.next("expr").value,
		body = ctx.next().value
	
	if (iterable && body.isBraces()) {
		return #`for ($_ of ${iterable}) ${body}` // yeah this wouldn't work because of hygiene
	}
}

each x of y {
	x()
}

each y {
	$_()
}
```

This would cover a lot of the same ground as readtables.

elibarzilay:
I'm not sure that I follow this completely, but it sounds fishy.  Specifically: "provided that they're used in different contexts" --- one major aspect of macros is that they are *local* transformations.

Also, I get the impression that this is kind of like allowing a function to return an "unknown" value, which could be handled by other definitions of the same function, so that one of them can return an answer.  But if this is true, then this smells like something that can suffer from an order-of-evaluation thing, and should anyway be addressed by proper code as you'd do in the case of functions.

Finally, I suspect that I'm missing something because of the last comment, since I don't see how this is related to readtables.

bopjesvla:
> I'm not sure that I follow this completely, but it sounds fishy. Specifically: "provided that they're used in different contexts" --- one major aspect of macros is that they are local transformations.

Yeah, that was sloppy wording. My use of "context" was colloquial there.  I meant "the usage of independently written macros with the same name in the same scope, provided they are used in different situations."

> Also, I get the impression that this is kind of like allowing a function to return an "unknown" value, which could be handled by other definitions of the same function, so that one of them can return an answer.

That's correct!

> But if this is true, then this smells like something that can suffer from an order-of-evaluation thing,

[...]

PR #552, opened 2016-06-05
Add reset and prev to macro context

gabejohnson:
This PR addresses #537.

The `reset` method simply reverts the macro context's enforester to the state it was in when the context was instantiated. This allows for many of the requests in #529.

I considered a few of the alternate naming suggestions, but none of them was a compelling alternative (though PHP does use `rewind`).

### DIFF
--- a/src/enforester.js
+++ b/src/enforester.js
@@ -332,7 +332,7 @@ export class Enforester {
     let lookahead = this.peek();
 
     if (this.term === null && this.isCompiletimeTransform(lookahead)) {
-      this.rest = this.expandMacro().concat(this.rest);
+      this.rest = this.expandMacro();
       lookahead = this.peek();
       this.term = null;
     }
@@ -1048,8 +1048,7 @@ export class Enforester {
     }
 
     if (this.term === null && this.isCompiletimeTransform(lookahead)) {
-      let result = this.expandMacro();
-      this.rest = result.concat(this.rest);
+      this.rest = this.expandMacro();
       return EXPR_LOOP_EXPANSION;
     }
 
@@ -1852,7 +1851,7 @@ export class Enforester {
       return stx.addScope(introducedScope, this.context.bindings, ALL_PHASES, { flip: true });
     });
 
-    return result;
+    return result.concat(ctx[Symbol.for('enforester')].rest);
 
   }
 
--- a/src/macro-context.js
+++ b/src/macro-context.js
@@ -10,6 +10,7 @@ const Nothing = Maybe.Nothing;
 
 const symWrap = Symbol('wrapper');
 const symName = Symbol('name');
+const symEnf = Symbol('enforester');
 const symResetValues = Symbol('resetValues');
 
 const getLineNumber = t => {
@@ -140,9 +141,7 @@ ctx :: {
 */
 export default class MacroContext {
   constructor(enf, name, context, useScope, introducedScope) {
-    // todo: perhaps replace with a symbol to keep mostly private?
-    this._enf = enf;
-    const { term, rest, prev, done} = enf;
+    const { term, rest, prev, done} = this[symEnf] = enf;
     this[symResetValues] = { term, rest, prev, done };
     this[symName] = name;
     this.context = context;
@@ -161,11 +160,11 @@ export default class MacroContext {
   }
 
   reset() {
-    Object.assign(this._enf, this[symResetValues]);
+    Object.assign(this[symEnf], this[symResetValues]);
   }
 
   next(type = 'Syntax') {
-    if (this._enf.rest.size === 0) {
+    if (this[symEnf].rest.size === 0) {
       return {
         done: true,
         value: null,
@@ -175,13 +174,13 @@ export default class MacroContext {
     switch(type) {
       case 'AssignmentExpression':
       case 'expr':
-        value = this._enf.enforestExpressionLoop();
+        value = this[symEnf].enforestExpressionLoop();
         break;
       case 'Expression':
-        value = this._enf.enforestExpression();
+        value = this[symEnf].enforestExpression();
         break;
       case 'Syntax':
-        value = this._enf.advance();
+        value = this[symEnf].advance();
         if (!this.noScopes) {
           value = value
             .addScope(this.useScope, this.context.bindings, ALL_PHASES)
--- a/src/macro-context.js
+++ b/src/macro-context.js
@@ -10,7 +10,7 @@ const Nothing = Maybe.Nothing;
 
 const symWrap = Symbol('wrapper');
 const symName = Symbol('name');
-const symEnf = Symbol('enforester');
+const symEnf = Symbol.for('enforester');
 const symResetVals = Symbol('resetVals');
 const symShadow = Symbol('shadow');
 
@@ -43,7 +43,7 @@ export class SyntaxOrTermWrapper {
       return stx.match(type, value);
     }
   }
-  
+
   isIdentifier(value) {
     return this.match("identifier", value);
   }
@@ -142,8 +142,11 @@ ctx :: {
 */
 export default class MacroContext {
   constructor(enf, name, context, useScope, introducedScope) {
-    const { term, rest, prev, done} = this[symEnf] = enf;
+    const { term, rest, prev, done } = this[symEnf] = enf;
+
     this[symResetVals] = { term, rest, prev, done };
+
+    // a shadow of the enforester stores previous terms to be used when calling prev()
     this[symShadow] = { term, rest, prev };
     this[symName] = name;
     this.context = context;
@@ -162,10 +165,9 @@ export default class MacroContext {
   }
 
   reset() {
-    let reset = this[symResetVals],
-        { term, prev, rest } = reset;
+    let { term, prev, rest } = this[symResetVals];
     Object.assign(this[symShadow], { term, prev, rest });
-    Object.assign(this[symEnf], reset);
+    this[symEnf] = new Enforester(rest, prev, this[symEnf].context);
   }
 
   next(type = 'Syntax') {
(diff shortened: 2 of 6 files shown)


gabejohnson:
I had a thought earlier today. We COULD provide a `rewind` method that optionally takes a number. No argument resets as is currently implemented. Providing an integer rewinds that number of iterations. Passing a number greater than the number of iterations throws and error.

Thoughts?

gabejohnson:
Another possibility is to just have `prev` w/ an object returned:
```js
{
  started: boolean,
  value: Syntax 
}
```

disnet:
So, I think what I like best at the moment is `prev` as the nice dual to `next` plus the convenience `reset`. This would extend nicely into infix macros once we implement them. For now `prev().done === true` when the iterator gets back to the start, but when we add infix it can just keep going for as long as we let it.

But there's a problem:

```js
syntax eat = ctx => {
  ctx.next();
  return #`1`;
}
syntax m = ctx => {
  ctx.next('expr'); // `eat 42` is expanded to `1` and then consumed
  ctx.next();       // consumes `;`
  ctx.prev();       // is this `42` or `1`???
}

m eat 42; 100
```

The original sin was complecting iteration with local expansion making it unclear if the iterator is over the original syntax or the modified. We can fix this with modules by making it clear that local expansion works on the output of iteration:

```js
import { localExpand } from 'sweet-helpers' for syntax;

syntax eat = ctx => {
  ctx.next();
  return #`1`;
}
syntax m = ctx => {
  let result = localExpand(ctx, 'expr');
  ctx.next();       // consumes `;`
  ctx.prev();       // obviously `42`
}

m eat 42; 100
```

So anyway, I guess what I'm saying is:

- `prev` + `reset` is good
- `prev` needs to work on the pre-expanded syntax
  - this might be surprising to users but we'll make this clear with a better API later

gabejohnson:
Do you want to merge the `reset` feature and then I'll open another issue for `prev`? Or just wait till I'm done w/ `prev`?

disnet:
Let's just do it all together.

gabejohnson:
@disnet: Should be ready to merge.

src/macro-context.js
```diff
@@ -184,9 +196,49 @@ export default class MacroContext {
...
     };
   }
+
+  prev() {
+    let shadow = this[symShadow];
+    let resetRestCount = this[symResetVals].rest.size;
+    if(resetRestCount > shadow.rest.size) {
+      let enf = this[symEnf];
+      Object.assign(shadow, recede(shadow));
+      Object.assign(this[symEnf], { term: null, rest: shadow.rest });
```
disnet: I really don't like the idea of mutating the enforester. Can you just create a new instance with `shadow.rest` as its argument?

src/macro-context.js
```diff
@@ -139,8 +142,9 @@ ctx :: {
 */
 export default class MacroContext {
   constructor(enf, name, context, useScope, introducedScope) {
-    // todo: perhaps replace with a symbol to keep mostly private?
-    this._enf = enf;
+    const { term, rest, prev, done} = this[symEnf] = enf;
```
disnet: formatting: `done}` -> `done }`

src/macro-context.js
gabejohnson: How do we keep the state of the original enforester up-to-date? Once the transform function returns, doesn't it's `rest` have to match `shadow.rest`?

gabejohnson: @disnet: Maybe the enforester could ask the macro context for the most up-to-date `rest` list after the transform returns?

src/macro-context.js
```diff
@@ -142,7 +142,8 @@ ctx :: {
 */
 export default class MacroContext {
   constructor(enf, name, context, useScope, introducedScope) {
-    const { term, rest, prev, done} = this[symEnf] = enf;
+    const { term, rest, prev, done } = enf;
+    this[symEnf] = new Enforester(rest, prev, enf.context);
```
gabejohnson: @disnet: This avoids mutating the enforester that was passed in. If you want I can do what you suggested and instantiate a new one each time `prev` is called.

src/enforester.js
```diff
@@ -1852,7 +1851,8 @@ export class Enforester {
       return stx.addScope(introducedScope, this.context.bindings, ALL_PHASES, { flip: true });
     });
 
-    return result;
+    // return result.concat(ctx._enf.rest);
+    return result.concat(ctx[Symbol.for('enforester')].rest);
```
gabejohnson: Referencing a global symbol feels a little dirty, but I'm struggling to find a better way.

src/enforester.js
```diff
@@ -1852,7 +1851,7 @@ export class Enforester {
       return stx.addScope(introducedScope, this.context.bindings, ALL_PHASES, { flip: true });
     });
 
-    return result;
+    return result.concat(ctx[Symbol.for('enforester')].rest);
```
gabejohnson: Referencing a global symbol feels a little dirty, but I'm struggling to find a better way.

disnet:
Gross. Yeah, sorry about my suggestion for creating new instances of `Enforester`. Much better to have the macro context mutate an enforester than have an enforester rely on a global symbol.

My concern with mutating the enforester is that it forces the mutating code to know about and maintain the enforester invariants. But a global symbol makes that tight coupling 10 times worse :/

Actually, maybe we can do even better! Rather than construct `MacroContext` with a reference to the enforester, just construct a `MacroContext` with a list of syntax objects. Then each call to `next`/`prev` in `MacroContext` can just keep an index into the list. Calls to `.next('expr')` can construct a new enforester by slicing the list and the calling `.enforest` (with the appropriate managing of the list index).

This way we never have to reset the invoking enforester because we never mutate it. More specifically, `expandMacro` in the enforester can ask its instance of `MacroContext` how much syntax was consumed during expansion and then mutate itself appropriately.

Does this make sense or am I forgetting something?

gabejohnson:
It makes sense and I was hoping you'd say something like that. Keeping an eye toward decomplecting enforestation from traversal (in the macro context) we should probably have some way to directly access the raw list in the context.

Another option is to add `.next('raw')` or some such so we can avoid adding context during traversal if needed.

In order to implement the `localExpand` helper function you referenced above we would need to create a new enforester using the raw list.

I could do that work as part of this PR or make another one and hold off on this PR until the other is complete.

If we're going to do away with `.next('expr')` and friends, we might as well do it now.

Thoughts?

gabejohnson:
@disnet another question is how to update the context's syntax list index if expression enforestation is being done in a helper? `ctx.step(n)`? `ctx.rest = stxl`?

disnet:
So I just relaized that separating out enforestation into `localExpand` doesn't actually work because enforestation/expansion requires access to the current expansion context (things like the binding map, current scopes, etc.). In addition, it raises gross questions like you raised about communicating how much syntax `localExpand` consumed back to the original macro context.

I spent the past few days thinking through alternate macro context designs and I think what I've come to is that traversal and enforestation are just fundamentally complected and that's ok. We could try to pretend they are not but then we just have to invent ways to connect them again.

So, the goal of the macro context should be to have a minimal and clear API.

Here's my proposal.

- add `.reset`
- drop `.prev`
- drop `.next('expr')`
- add `.expand('expr')`

I think we should not add `.prev` to the macro context since that is something that could be implemented as a library (and as of Sweet 1.1's import `for syntax` we can start doing that now).

We should move the enforestation functionality to `.expand` to clarify its doing something very different than simply getting the next expression (in particular, expansion can have side effects). If we can't decomplect entirely we can at least do so at the API naming level.

`.reset` is essential because we need a way to "undo" the expansion inside `.expand`

Anyway, does this seem reasonable? Sorry about the back and forth, design is hard.

gabejohnson:
If it wasn't difficult, it wouldn't be fun :wink:

I've also spent the last couple of days trying to separate traversal from enforestation. I tried passing the context AND syntax explicitly into the transform function. There was a problem in the term expander that I couldn't track down but realized that while it was adding flexibility, the complexity for the macro author probably wouldn't be worth it. But it's very tempting to have a handle on that syntax list.

Your new API looks reasonable, but there's still the issue of communicating back to the enforester how much syntax has been consumed (assuming no direct mutation by the macro context). Should we have `MacroContext.prototype.syntaxRemaining`, `.syntaxConsumed`? A simpler option would be to require that the rest of the syntax be returned from the transform and not have the enforester concat the result of the transform with the rest of the syntax list:

```js
syntax pair = (ctx) => {
  let fst = ctx.next().value;
  let snd = ctx.next().value;
  return #`[${fst}, ${snd}] ${ctx}`; // [30, 42], 66
}

pair 30, 42, 66;
```
another (less aesthetically pleasing) option:

```js
return [#`[${fst}, ${snd}]`, [...ctx]];
```

I don't think forcing the macro author to think about the entire local context in which the transform is occurring is necessarily a bad thing. One could argue that it just introduces boilerplate, but considering most authors are going to be using something like `syntaxCase` and `syntaxRules`, it shouldn't be much of an issue. To my mind, `syntax` and `syntaxrec` are pretty low-level.

disnet:
> Should we have `MacroContext.prototype.syntaxRemaining`, `.syntaxConsumed`?

I think I like this one best. The boilerplate in a transformer wouldn't be a problem (`syntaxCase` will eventually save us) but it would need to completely consume the iterator to get the remaining syntax, which is pointless work.

Let's make it clear it is private, so maybe call it `MacroContext::_rest` and lets actually enforce its privacy. I don't want a syntax transformer to gain direct access to the unmediated syntax list.

Maybe something like a keyed predicate?

```js
  expandMacro() {
    //...
    let KEY = {};
    let ctx = new MacroContext(/* ... */, x => x === KEY);

    // ...

    let rest = ctx._rest(KEY);  // _rest uses the predicate from the constructor
  }
```

A weak map might work as well.

gabejohnson:
Okay. I'll try the weak map strategy first as it won't change the signature of the `MacroContext` constructor.

```js
let rest = ctx._rest(this);
```

What about the behavior of `.expand` when the syntax list is exhausted?

disnet:
Make the return type of `expand` the same as `next`.

Just say no to `null` 😄

gabejohnson:
Will do. Do you want me to submit a new PR? This new implementation doesn't really have much in common with the old.

disnet:
Good idea.

PR #556, opened 2016-06-20
Add reset and expand to MacroContext

gabejohnson:
Re-submit of #552.

This PR adds `.reset` which allows the context to be returned to its initial state. It also removes the enforestation logic from `.next` to `.expand`.

### DIFF
--- a/src/macro-context.js
+++ b/src/macro-context.js
@@ -163,7 +163,10 @@ export default class MacroContext {
   expand(type) {
     const enf = privateData.get(this).enf;
     if (enf.rest.size === 0) {
-      return null;
+      return {
+        done: true,
+        value: null
+      };
     }
     let value;
     switch(type) {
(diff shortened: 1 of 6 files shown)


src/macro-context.js
```diff
@@ -139,8 +140,10 @@ ctx :: {
 */
 export default class MacroContext {
   constructor(enf, name, context, useScope, introducedScope) {
-    // todo: perhaps replace with a symbol to keep mostly private?
-    this._enf = enf;
+    privateData.set(this, { backup: enf });
+    this.reset(); // instantiate enforester
+
+    //TODO: add more fields to privateData?
```
disnet: Can you hide `context`, `useScope`, and `introducedScope` in the weak map too?

gabejohnson: I just threw all the members in there. Couldn't find a reason not to. Is that cool?

disnet:
This is looking great! Unless you had anything else you wanted to do I'll go ahead and merge it.

gabejohnson:
Nothing I can think of. Merge away.

PR #557, opened 2016-06-22
Properly enforest Left-Hand-Side Expressions

### LINKED ISSUE #548: Sweet.js can't parse expression with `new` if it's not followed by an identifier
1)
```js
var instance = new(Function.prototype.bind.apply(Date, [null].concat(args)))();
```
Leads to `Error: [assertion error]: cannot get the val of a delimiter`

2)
```js
new (Date)
```
leads to `Error: expecting an identifier`

gabejohnson:
fixes #548


disnet:
This doesn't completely solve the problem. We actually need to be matching a `LeftHandSideExpression` there (e.g. `new [1, 2, 3]` should also be valid).

I'm fine with merging this if it unblocks you and making the general fix later (since this PR makes things strictly better) or you could tackle implementing `enforestLeftHandSideExpression` now.

gabejohnson:
I can work on that. Are you thinking in the manner of `enforestPrimaryExpression` from #558?

disnet:
Yeah, similar to that. You might want to refer to what the shift parser [does](https://github.com/shapesecurity/shift-parser-js/blob/2e34679437e5c24abecc03e6d67d4240d3fe0d27/src/parser.js#L1399-L1479).

gabejohnson:
@disnet: I'm a little foggy on why `new [1, 2, 3]` should parse as an ArrayExpression following `new` doesn't make sense..

Should a LeftHandSideExpression be able to be assigned an arbitrary expression? I thought it would only accept a constructor.

gabejohnson:
Or is the expression not resolved until runtime?

disnet:
Right, it's not resolved until runtime (at which point it throws a `TypeError`) but it's still a "Program" so we have to parse it as such.

gabejohnson:
@disnet: Can you look this over? I think it's a go.

src/enforester.js
```diff
@@ -1314,17 +1327,26 @@ export class Enforester {
...
       return new Term('NewTargetExpression', {});
     } else {
       callee = new Term('IdentifierExpression', { name : this.enforestIdentifier() });
     }
+
+    if(this.isTemplate(this.peek())) {
+      this.term = callee;
+      callee = this.enforestLeftHandSideExpression({ allowCall: false });
+    }
+
```
disnet: Besides `new.target` all these cases can be handled in `enforestLeftHandSideExpression` right?

gabejohnson: It's really just pushing around the complexity IMO. It places a lot of `new` specific code into `enforestLeftHandSideExpression`. I can try to move it if you want, but I don't see what's gained.

src/enforester.js
```diff
@@ -1314,17 +1327,26 @@ export class Enforester {
...
     let callee;
-    if (this.isKeyword(this.peek(), 'new')) {
+    let lookahead = this.peek();
+    if (this.isKeyword(lookahead, 'new')) {
       callee = this.enforestNewExpression();
-    } else if (this.isKeyword(this.peek(), 'super')) {
+    } else if (this.isKeyword(lookahead, 'super')) {
+      callee = this.enforestLeftHandSideExpression({ allowCall: false });
+    } else if (this.isParens(lookahead) || this.isBrackets(lookahead) || this.isBraces(lookahead) || this.isTemplate(lookahead)) {
       callee = this.enforestExpressionLoop();
```
gabejohnson: I can move the logic in the above four lines to `enforestLeftHandSideExpression`. Everything else handles `new` specific edge cases.

src/enforester.js
```diff
@@ -1314,17 +1327,26 @@ export class Enforester {
   enforestNewExpression() {
     this.matchKeyword('new');
     let callee;
-    if (this.isKeyword(this.peek(), 'new')) {
+    let lookahead = this.peek();
+    if (this.isKeyword(lookahead, 'new')) {
       callee = this.enforestNewExpression();
```
gabejohnson: This one too I guess. Okay. You've convinced me!

src/enforester.js
```diff
@@ -1326,25 +1328,20 @@ export class Enforester {
...
-      callee = new Term('IdentifierExpression', { name : this.enforestIdentifier() });
+      this.term = this.enforestLeftHandSideExpression({ allowCall: false });
     }
 
+    // (new a``) is a TemplateExpression while (new a()``) is a NewExpression
     if(this.isTemplate(this.peek())) {
-      this.term = callee;
-      callee = this.enforestLeftHandSideExpression({ allowCall: false });
+      this.term = this.enforestLeftHandSideExpression({ allowCall: false });
     }
```
gabejohnson: @disnet: is this better?

src/enforester.js
```diff
@@ -1313,26 +1328,30 @@ export class Enforester {
...
+      this.term = new Term('IdentifierExpression', { name : this.enforestIdentifier() });
     } else {
-      callee = new Term('IdentifierExpression', { name : this.enforestIdentifier() });
+      this.term = this.enforestLeftHandSideExpression({ allowCall: false });
+    }
+
+    // (new a``) is a NewExpression while (new a()``) is a TemplateExpression
+    if(this.isTemplate(this.peek())) {
+      this.term = this.enforestLeftHandSideExpression({ allowCall: false });
     }
```
gabejohnson: @disnet: is that better?

src/enforester.js
```diff
@@ -1326,25 +1328,20 @@ export class Enforester {
...
-      callee = new Term('IdentifierExpression', { name : this.enforestIdentifier() });
+      this.term = this.enforestLeftHandSideExpression({ allowCall: false });
     }
 
+    // (new a``) is a NewExpression while (new a()``) is a TemplateExpression
     if(this.isTemplate(this.peek())) {
-      this.term = callee;
-      callee = this.enforestLeftHandSideExpression({ allowCall: false });
+      this.term = this.enforestLeftHandSideExpression({ allowCall: false });
     }
```
gabejohnson: @disnet: You were right. It's cleaner. I think I'm still stuck with 1343-1345 though, unless you have some ideas. I guess I should hold off on responding until I have my :coffee:

disnet: I'm not following why you need separate cases for identifier expressions and templates. Aren't these both handled by `enforestLeftHandSideExpression`?

btw, the main reason I suggested moving the cases to `enforestLHSE` is that I think we actually need to use `enforestLHSE` in a few other places (e.g. class extends clause) so it should be as general as we need. We can fix up those other cases once this PR lands; we just need the fully general `enforestLHSE`.

gabejohnson: Sorry about that. I was perceiving a special case for new expressions. I'm about to push code that early aborts `enforestLHSE` if `this.term` is an `IdentifierExpression` and calls are not allowed.

src/enforester.js
```diff
@@ -1313,26 +1337,23 @@ export class Enforester {
...
-      callee = this.enforestExpressionLoop();
-    } else if (this.isPunctuator(this.peek(), '.') && this.isIdentifier(this.peek(1), 'target')) {
+    let lookahead = this.peek();
+    if (this.isPunctuator(lookahead, '.') && this.isIdentifier(this.peek(1), 'target')) {
       this.advance();
       this.advance();
       return new Term('NewTargetExpression', {});
     } else {
-      callee = new Term('IdentifierExpression', { name : this.enforestIdentifier() });
+      this.term = this.enforestLeftHandSideExpression({ allowCall: false });
```
disnet: I don't think you need to assign to `this.term` here, the local `callee` should work better.

gabejohnson: Yeah. I think that's an artifact of my earlier efforts. I'll fix it ASAP.

disnet:
I merged this. Squashed so I guess github didn't notice :)

Thanks!

PR #564, opened 2016-08-01
Fix Ex 5 in the tutorial to match its expansion

felipethome:
The expansion of the example 5 in the tutorial, with sweet.js version 2.0.0, gives for me:

    (function (bb8_5) {
      ;
      console.log(bb8_5.beep());
    }(new Droid("BB-8", "orange")));

So I added one more ctx.next() and rebuild the docs. Maybe would be better also a check to see if the `;` exists?

### DIFF
--- a/doc/1.0/tutorial.adoc
+++ b/doc/1.0/tutorial.adoc
@@ -157,6 +157,7 @@ syntax let = function (ctx) {
   let ident = ctx.next().value;
   ctx.next(); // eat `=`
   let init = ctx.expand('expr').value; // <1>
+  ctx.next(); // eat `;`
   return #`
     (function (${ident}) {
       ${ctx} // <2>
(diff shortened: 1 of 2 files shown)


disnet:
Right, you would need to check to make sure the `;` exists for the example to really work but that just makes the example more confusing than necessary. Better to have a harmless empty statement in the expansion output imo.

felipethome:
But don't you think it is nice for the user to see `;` needs to be eaten too? Also, all the examples have their corresponding expansion, but in this particular case the expansion is misleading because it doesn't contain the `;`. If you disagree feel free to close it @disnet, and thanks for the attention.

disnet:
Actually, why don't you remove the `;` from the source? That demonstrates that expansion is ASI aware but avoids complicating the macro code.

PR #565, opened 2016-08-05
Add the ability to create a reset point on a macro context

gabejohnson:
The ability to move back one term can be implemented as a utility function. Unfortunately the time complexity is O(n) in the number of syntax items.

To my mind a big reason to have this ability is to expand a term based on the type of syntax read:
```js
import { prev } from './utils' for syntax;

syntax m = ctx => {
  ctx.next() // eatThis
  let stx = ctx.next().value;
  let ret;
  if(stx.isIdentifier()) {
    ret = #`${stx}`;
  } else if(stx.isDelimiter() && stx.isBracket()) {
    prev(ctx); // <- call to prev
    ret = #`${ctx.expand('ArrayExpression')}`;
  }
  return ret;
}

m eatThis foo; // foo
m eatThis [1,2,3] // [1,2,3]
```

But the same thing can be accomplished with lookahead:
```js
import { lookahead } from './utils' for syntax;

syntax m = ctx => {
  ctx.next(); // eatThis
  let stx = lookahead(ctx); // <- call to lookahead
  let ret;
  if(stx.isIdentifier()) {
    ret = #`${stx}`;
  } else if(stx.isDelimiter() && stx.isBracket()) {
    ret = #`${ctx.expand('ArrayExpression')}`;
  }
  return ret;
}

m eatThis foo; // foo
m eatThis [1,2,3] // [1,2,3]
```
With this PR `lookahead(1)` can be implemented in constant time and `lookahead(m)` is O(m).

The feature works as follows:
```js
syntax m = ctx => {
  ctx.next(); // eatThis
  ctx.mark(); // <- marks a reset point
  let stx = ctx.next().value
  let ret;
  if(stx.isIdentifier()) {
    ret = #`${stx}`;
  } else if(stx.isDelimiter() && stx.isBracket()) {
    ctx.reset(); // <- back to reset point
    ret = #`${ctx.expand('ArrayExpression')}`;
  }
  return ret;
}

m foo; // foo
m [1,2,3] // [1,2,3]
```

### DIFF
--- a/src/macro-context.js
+++ b/src/macro-context.js
@@ -306,7 +306,7 @@ export default class MacroContext {
 
   _rest(enf) {
     const priv = privateData.get(this);
-    if(priv.markers.get(priv.startMarker) === enf) {
+    if (priv.markers.get(priv.startMarker) === enf) {
       return priv.enf.rest;
     }
     throw Error("Unauthorized access!");
@@ -315,12 +315,12 @@ export default class MacroContext {
   reset(marker) {
     const priv = privateData.get(this);
     let enf;
-    if(marker == null) {
+    if (marker == null) {
       // go to the beginning
       enf = priv.markers.get(priv.startMarker);
-    } else if(marker && marker instanceof Marker) {
+    } else if (marker && marker instanceof Marker) {
       // marker could be from another context
-      if(priv.markers.has(marker)) {
+      if (priv.markers.has(marker)) {
         enf = priv.markers.get(marker);
       } else {
         throw new Error('marker must originate from this context');
@@ -334,16 +334,20 @@ export default class MacroContext {
   mark() {
     const priv = privateData.get(this);
     let marker;
-    if(priv.enf.rest.size === priv.markers.get(priv.startMarker).rest.size) {
+
+    // the idea here is that marking at the beginning shouldn't happen more than once.
+    // We can reuse startMarker.
+    if (priv.enf.rest.size === priv.markers.get(priv.startMarker).rest.size) {
       marker = priv.startMarker;
-    } else if(priv.enf.rest.isEmpty()) {
-      if(!priv.endMarker) priv.endMarker = new Marker();
+    } else if (priv.enf.rest.isEmpty()) {
+      // same reason as above
+      if (!priv.endMarker) priv.endMarker = new Marker();
       marker = priv.endMarker;
     } else {
       //TODO(optimization/dubious): check that there isn't already a marker for this index?
       marker = new Marker();
     }
-    if(!priv.markers.has(marker)) {
+    if (!priv.markers.has(marker)) {
       priv.markers.set(marker, cloneEnforester(priv.enf));
     }
     return marker;
--- a/src/macro-context.js
+++ b/src/macro-context.js
@@ -172,6 +172,13 @@ export function unwrap(x) {
   return x;
 }
 
+function cloneEnforester(enf) {
+  const { rest, prev, context } = enf;
+  return new Enforester(rest, prev, context);
+}
+
+function Marker () {}
+
 /*
 ctx :: {
   of: (Syntax) -> ctx
@@ -179,11 +186,15 @@ ctx :: {
 }
 */
 export default class MacroContext {
+
   constructor(enf, name, context, useScope, introducedScope) {
+    const startMarker = new Marker();
+    const startEnf = cloneEnforester(enf);
     const priv = {
-      backups: List.of(enf),
       name,
-      context
+      context,
+      startMarker,
+      markers: new Map([[startMarker, enf]]),
     };
 
     if (useScope && introducedScope) {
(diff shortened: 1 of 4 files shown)


gabejohnson:
@disnet if you give the go-ahead, I'll update the docs.

disnet:
👍 

Does it make sense for `mark` to return a mark pointer and `reset` to take a mark pointer? That way you could have multiple marks to reset back to?

vendethiel:
How do you track which point are after which points? Or is it okay to save a, do stuff, save b, do stuff, load a, load b?

gabejohnson:
@disnet I guess that could be useful for returning to a particular point repeatedly. Do you have a particular use case in mind?

@vendethiel Just what you described. If you wanted to ensure order you could store them in a sequence.

disnet:
I'm thinking of marks as branches in a decision tree. You might be deep in the tree and want to reset all the way to the top (first mark).

gabejohnson:
Without having to track the number marks you've made. Yeah, I can dig it. I'll put something together a little bit later.

What about `reset` no arg behavior? The same as before (to the beginning) or closest mark behind current position?

I'm a little torn. Automatically resetting to the beginning is convenient, but so is just going back to the last breadcrumb.

gabejohnson:
On second thought, you could do what I suggested to vendethiel if you just want to go back one.

disnet:
> What about `reset` no arg behavior? 

Definitely should go back to the beginning.

gabejohnson:
@disnet do you want me to update the docs?

gabejohnson:
Do you want `ctx.reset(marker)` to throw if the `marker` is from another context?

disnet:
Yes to updating docs and throwing on the wrong marker (or non-marker values).

gabejohnson:
@disnet Does this look good to you?

src/macro-context.js
```diff
@@ -172,18 +172,29 @@ export function unwrap(x) {
...
   constructor(enf, name, context, useScope, introducedScope) {
+    const startMarker = new Marker();
+    const startEnf = cloneEnforester(enf);
     const priv = {
-      backup: enf,
       name,
-      context
+      context,
+      startMarker,
+      markers: new Map([[startMarker, enf]]),
```
disnet: should this be a `WeakMap`?

src/macro-context.js
```diff
@@ -295,16 +306,47 @@ export default class MacroContext {
...
       return priv.enf.rest;
     }
     throw Error("Unauthorized access!");
   }
 
-  reset() {
+  reset(marker) {
+    const priv = privateData.get(this);
+    let enf;
+    if(marker == null) {
```
disnet: formatting: `if(...` -> `if (...`

src/macro-context.js
```diff
@@ -295,16 +306,47 @@ export default class MacroContext {
...
+    }
+    priv.enf = cloneEnforester(enf);
+  }
+
+  mark() {
     const priv = privateData.get(this);
-    const { rest, prev, context } = priv.backup;
-    priv.enf = new Enforester(rest, prev, context);
+    let marker;
+    if(priv.enf.rest.size === priv.markers.get(priv.startMarker).rest.size) {
```
disnet: It's unclear at first glance what this first condition is doing. I *think* it's to handle marking at the very beginning but it's not clear why we have to compare sizes. Perhaps we can store a reference to the first `priv` and just check for that?

src/macro-context.js
```diff
@@ -295,16 +306,47 @@ export default class MacroContext {
...
+      marker = priv.startMarker;
+    } else if(priv.enf.rest.isEmpty()) {
+      if(!priv.endMarker) priv.endMarker = new Marker();
```
disnet: Why store an `endMarker`? Seems like it's not used.

src/macro-context.js
gabejohnson: I didn't think `WeakMap` necessary as the `markers` doesn't outlive the `MacroContext` in which it's contained.

src/macro-context.js
gabejohnson: Do you mean store a reference to the first `enf`? That is what `startMarker` is the key for. I should add a comment, but the idea is that marking at the beginning shouldn't happen more than once.
The reason I can't just compare `enf` and `markers.get(startMarker)` is that they aren't the same enforester. I guess I could compare the lists and not the size.

src/macro-context.js
gabejohnson: It _is_ being used on the next line. Again, the idea is marker reuse.

src/macro-context.js
gabejohnson: Yeah. I need to turn on a linter. I'll go through the file and fix them all.

gabejohnson: Turns out I did have my linter on. Do you want me to add the rule to `.eslintrc`?

disnet: Yeah, that'd be great!

src/macro-context.js
disnet: Ah, that makes sense. A comment should be fine then.

src/macro-context.js
disnet: haha got confused, that does make sense.

disnet:
Thanks!

PR #571, opened 2016-08-17
Small typo fix

gnclmorais:
It’s basically `framwork -> framework`. I think I only needed to change the .adoc file, but I was not 100% sure (and I couldn’t find a _contributing.md_ file on this repo), so I just went ahead.


disnet:
Thanks!

PR #593, opened 2016-08-31
fixes #592

### LINKED ISSUE #592: `new foo.bar()` is not compiled correctly
Without using any macros, sweet.js outputs the wrong code for this:

```js
const db = new sqlite.Database("foo")
```

The generated output is:

```js
const db_0 = new "foo";
```

gabejohnson:
This fix handles new expressions with member expressions as callees.


disnet:
Thanks!

Issue #535, opened 2016-04-16
Add Readtables back


elibarzilay:
Again, apologies for the vague issue, but:
1. Do readtables exist?  If so where are they documented?  (Better: some simple examples?)
2. If not, then make this a feature request to have them...

disnet:
Ah, thought I had already made an issue for this.

Indeed, they are missing and we need to add them back. We had them pre-1.0 (thanks @jlongster!) but then I rewrote the reader.

ericguedespinto:
Could we use readtables to make sweetjs whitespace aware? Say for instance like coffescript or ruby?

vendethiel:
That'd be really, really, really hard.

elibarzilay:
Readtables are more intended for a case where you want to add a new construct to an existing language with minimal changes.  An indentation-based syntax is a drastic change to the language so you really need to implement a parser for that.

bopjesvla:
I [implemented](https://gist.github.com/bopjesvla/768b1bdc99a0f97828273ec4146e1892) indent blocks in 0.7. The table used a simple regular expression to track the flow of indentation. The syntax was:

```
while (true):
	while (false):
		if (1 === 2):
			return;
		y = 3
		j = 5
	x = 5
	for (i of [3]):
		i.j="k"
	i = 3
```

Unfortunately, outdenting two levels at once was blocked by an issue related to nested braces, IIRC.

gabejohnson:
@disnet are you looking to have the same API as the old readtables?

disnet:
Not necessarily.

gabejohnson:
I'm looking to extend the utility of `#`. I'm thinking of looking at Racket's API for inspiration. What requirements are you targeting for readtables?

gabejohnson:
So far I've come up with a primitive api that's similar to that of `syntax`:
```js
lexeme ~> = function(reader) {
  let result;
  const key = reader.key(); // ~>
  reader.read(); // eats 'skipMe'. uses default reader behavior
  // creates a reset point that is consumed when reset is called.
  // other possible names: shift, savePos, checkpoint
  reader.mark();
  const delim = reader.next().value // [
  let brackets;
  if (delim.isLeftBracket()) {
    reader.reset();
    brackets = reader.read('Delimiter');
  } else { // throw }
  return `thread { ${brackets.inner()} }`; // .inner returns another reader instance
}

foo ~> skipMe [bar, baz(1)] // foo thread { bar, baz(1) }
```
[...]

PR #601, opened 2016-09-14
Implement tokenization via readtables

gabejohnson:
Note: This work is definitely not complete, functional or ready to be merged.

This is my first (published) pass at a readtable implementation. It isn't hooked into the pipeline yet but does show the general strategy and initial API. Per some previous discussions https://github.com/sweet-js/sweet.js/issues/535#issuecomment-242791944 I've intentionally implemented a very limited API (utility libs can build on these).

Readtables can currently be created as follows:

``` js
import { defaultReadtable } from 'src/readtable';
import CharStream from 'src/char-stream';

const newTable = defaultReadtable.extendReadtable({
  key: '@',
  action(stream) {
    return {
      type: 'Punctuator',
      value: stream.readString(),
      locationInfo: stream.locationInfo
    };
  }
});

const entry = newTable.getEntry('@');
const stream = new CharStream('@foo');
entry.action(stream);
// =>
// {
//   type: 'Punctuator',
//   value: '@',
//   locationInfo: {
//     filename: '',
//     line: 0,
//     column: 0,
//     position: 0
//   }
// }
```

The intention is to allow lookbehind of already read tokens. I'll have to implement this for '/' resolution. So the signature of `action` will be `action(stream: CharStream, prefix: TokenTree?)`. `TokenTree = Token | List<TokenTree>`.

I have not implemented Racket style reader modes as I believe they can be build using the current primitives (i.e. by referencing other readtables in a closure).

As a note, I'm using some "private" fields (denoted by a `_` prefix). I'm reserving the right to truly hide them in the future if there is a reason and efficiencies don't suffer too greatly. So their presence shouldn't be relied upon and shouldn't be part of the public API.

@disnet I'd appreciate some feedback as soon as you get the chance. I'm going to have a lot of time the next few days to hack on this and don't want to go too far down the wrong path.


disnet:
This is looking awesome!

One thing I'm unclear on is how you would read a token from inside an action (dispatching to the current readtable). I think in our previous discussion I suggested putting a `read` method on `CharStream` but it might make more sense to put it on the `this` object in an action. Thoughts?

gabejohnson:
Yeah. Right now you can do one of two things:

1. reference a readtable you've created
2. reference the current readtable in `action`.

`CurrentReadtable` is an export from `readtable.js`, but I have a note in there to move it. There's also `setCurrentReadtable`. Another possibility is to access the current readtable procedurally.

I'll have to think about `this` in `action`. We could certainly have a `read` method on the reader and bind the reader to all calls to `action`. I have a current (possibly passing) bias against using `this` unless I'm forced to. I actually went so far as to implementing all of the methods on `Readtable` and `CharStream` as exported functions. Just changed that this morning.

But this IS JavaScript after all! 😉

gabejohnson:
A point in `this`'s favor: `prefix` could be accessed directly from the reader and not passed in to `action`.

gabejohnson:
So maybe this:
```js
function action(stream) {
  const { readtable, prefix } = this;
  const char = stream.readString();
  // inspect prefix ...
  const action = readtable.getEntry(char);
  return action.call(this, stream);
}
```
But then I'm jumping through hoops to avoid putting `read` on `this`.

I'll see where it goes as I add delimiters to the default readtable.

disnet:
Yeah, the key is you need to late-bind on whatever the current readtable is. I hear OO languages have this (lol) crazy feature to do late-binding :P

gabejohnson:
@disnet I'm getting pretty close now. Just have to do syntax templates (should be trivial) and then wire everything in.

One thing I'm not clear on is how to use `read` function with the module context.  Could we add the context during a later pass? Or do you think just passing it as an optional parameter to `read`?

disnet:
> One thing I'm not clear on is how to use read function with the module context. 

I'm not following. What do you mean by module context?

gabejohnson:
@disnet I guess it's really the whole program context. It's passed when creating a `Modules` instance. I decided to just pass it as an optional parameter to `read`.

I finished syntax templates but realized that I forgot about comments and `#lang`. I'm going to punt on `#lang` for now (treat it like a comment).

On side note, I'm not finding an tests for the replacement values passed to the current `Reader` from a `syntaxQuote`. What are the interpolated values supposed to be? Looking at the `Reader` source it appears they're tokens.

disnet:
> Just the bindings and scopesets are needed by the reader so I decided to just pass it as an optional parameter to read.

Ah yeah that's fine. 

> I'm going to punt on #lang for now (treat it like a comment).

Yep, we can work though how `#lang` should be handled later.

>  I'm not finding an tests for the replacement values passed to the current Reader from a syntaxQuote.

I actually want to remove `syntaxQuote` (and I'm doing that in #602) so I think you can just not worry about it.

gabejohnson:
Everything is wired together and tests are passing! I'm going to try to unify error handling and do some cleanup and reorg. The last thing is to get location info working properly.

I'd love it if you'd give it a glance for any glaring stupidities 😄

test/test_default_readtable.js
```diff
@@ -59,7 +61,7 @@ test('should parse keywords', t => {
     });
   }
 
-  testParseKeyword('await ', 'await');
+  // testParseKeyword('await ', 'await'); TODO: uncomment when await is a keyword
```
disnet: is there some reason you are waiting to make `await` a keyword?

gabejohnson: It's not a keyword in ES2015-16. Test fails at https://github.com/sweet-js/sweet.js/blob/22e5d182eb764d8427a7f447a69ad08ab777cb7a/test/expressions/identifier-expression.js#L31.
I could comment out that test and add `async` as well, but I figured I'd wait until we were using `babel-preset-es2017`. That would be after an upgrade to the Shift suite I guess.

gabejohnson: Of course there's a difference 😒

src/tokens.js
```diff
@@ -0,0 +1,279 @@
+// @flow
+
+import type { List } from 'immutable';
+
+import type Syntax from './syntax';
+
+import { TokenType, TokenClass } from 'shift-parser/dist/tokenizer';
```
disnet: is it possible to remove this dependency and use our own `TokenType` and `TokenClass` objects?

src/default-readtable.js
```diff
@@ -0,0 +1,231 @@
...
+
+const lineTerminatorEntries = lineTerminatorTable.map(lt => ({
+  key: lt,
+  action: function readLineTerminator(stream) {
+    this.locationInfo = { line: this.locationInfo.line + 1, column: 1 };
+    return eatWhitespace(stream);
+  }
+}));
+
+const digits = ['0', '1', '2', '3', '4', '5', '6', '7', '8'];
```
gabejohnson: Wow. Nice catch! I'll add a test.

gabejohnson:
@disnet I'm seeings ~60% increase in speed (assuming it isn't due to a bug) over `shift-reader` using https://github.com/shapesecurity/shift-parser-js/blob/ce88859a4636ba1d0d06d7d3e0d89ade767c8233/profile.js as a starting point.

I think it's due to the method used in readtable lookup in addition to how punctuators and keywords are resolved.

gabejohnson:
How were you thinking we'd make the`read`, `getCurrentReadtable` and `setCurrentReadtable` available to users? As imports or as globals?

disnet:
I think imports make the most sense right?

gabejohnson:
I agree, just wanted to make sure you didn't want `read` to be like `eval`.

gabejohnson:
I think this is finally ready to go. Once we get this and #602 merged I'll see about adding modes and updating the docs.

Just to be clear, we are _not_ exposing this functionality to users for the moment. How are we going to do that?

disnet:
> Just to be clear, we are not exposing this functionality to users for the moment. How are we going to do that?

With the `#lang` pragma. I'd actually like to use directives instead of our current syntax. So, a module that wants to opt into a language with a custom reader would look something like:

```js
'lang ./foo.js';

// ...
```

Following Racket, the `foo.js` module exports a `read` function that Sweet can use instead of the default `read`:

```js
import { read as readOrig } from 'sweet/reader/token-reader';
import { getCurrentReadtable, setCurrentReadtable } from 'sweet/reader/reader';

export function read(source, context) {
  let myReadtable = getCurrentReadtabl.extendReadtable(/* ... */);
  setCurrentReadtable(myReadtable);
  return readOrig(source, context);
}
```

One thing I'm not sure how we should handle is the current readtable modification. The problem is that Sweet needs to be careful about restoring the readtable to its original state after the `foo.js`'s `read` function has finished. Racket approaches this with `parameterize`, a kind of dynamic-scoping feature they have.

src/reader/default-readtable.js
```diff
@@ -0,0 +1,242 @@
...
+// 3. numbers - check!
+// 4. identifiers - check!
+// 5. string - check!
+// 6. regex - check!
+// 7. templates - check!
+// 8. keywords - check!
+// 9. dispatch characters (e.g. #, @) - check!
+// 10. '`' dispatch macro (i.e. syntaxQuote/syntaxTemplate) - check!
+
+// use https://github.com/mathiasbynens/regenerate to generate the Unicode code points
```
disnet: might want to remove your TODOs :)

test/test_char_stream.js
```diff
@@ -0,0 +1,41 @@
...
+
+// test('readChar should consume the next character', t => {
+//   t.is(stream.readChar(), 'a');
+//   t.is(stream.peek(), 'b');
+// });
+
+// test('readChar should return eos if it reaches the end of source', t => {
+//   readChar(stream); readChar(stream); readChar(stream);
+//   t.true(isEOS(readChar(stream)));
+// });
```
disnet: these should pass right? any reason to comment them out?

src/load-syntax.js
```diff
@@ -78,8 +78,7 @@ export function evalCompiletimeValue(expr, context) {
   let sandbox = {
     syntaxQuote: function (strings, ...values) {
       let ctx = deserializer.read(_.last(values));
-      let reader = new Reader(strings, ctx, _.take(values.length - 1, values));
-      return reader.read();
+      return read(strings.join(''), ctx)//reader.read();
```
disnet: clean extraneous comment

gabejohnson:
> With the #lang pragma. I'd actually like to use directives instead of our current syntax. So, a module that wants to opt into a language with a custom reader would look something like

I'm thinking about this in the context of composable languages (perhaps prematurely). Would something like this make sense then?
```js
'lang foo.js, bar.js, ./baz.js';
// ...
{ 'lang sweet.js'
// ...
}
```
The sweet.js language would be scoped to the enclosing block.

> Following Racket, the foo.js module exports a read function that Sweet can use instead of the default read

Now I was thinking we would just default export a readtable and maybe then the entries as additional exports. That way the module could provide both an entire extended language and component entries suitable for composition. Just thinking out loud.

But to my point. Are we just not going to document the new features for now? Maybe warn that they are there but caution them that they are experimental and can change w/o warning? Authors will still be able to import them.

disnet:
> I'm thinking about this in the context of composable languages (perhaps prematurely).

Block scoped language are I think straightforward to add. Just need a special form akin to Racket's `(module lang ...)`. Maybe `module 'lang ./foo.js' { ... }`.

For language "mixins" I think the directive syntax gives us some space to experiment with down the road. For example, we could use `'lang #mixin foo.js, ./bar.js baz'` to mean compose the readtables exported from the three modules in some clever way. 

> I could see the benefit of exporting a read function though. If the language wanted to set up something more extensive before tokenization (e.g. set up additional dispatch tables, use an entirely different subclass of Reader).

Exactly, for now `read` give the most flexibility. You can use the readtable if you want or do some other crazy thing if you need to. From the expander's perspective all that matters is that text turns into tokens.

Down the road we can provide forms of the language directive like `'lang #mixin ...'` that require a readtable (or some other even more composable structure we invent) to be exported. 

> But to my point. Are we just not going to document the new features for now?

Ah right. Yeah let's hold off on documentation. I'll finish reviewing, get this and #602 merged, and then wire up the language directive. From there we can make sure everything is working as we thought it would and then write up docs and spread the good news.

gabejohnson:
Sounds good. Once #602 is merged I'll start experimenting with modes.

vendethiel:
Really nice work, happy to see it's being done :).

disnet:
This is really nice code. Thanks for all your hard work!

src/reader/read-identifier.js
```diff
@@ -0,0 +1,92 @@
...
+      }
+    } else if (0xD800 <= code && code <= 0xDBFF) {
+      if (isEOS(char)) {
+        throw this.createILLEGAL(char);
+      }
+      let lowSurrogateCode = sPeek(1).charCodeAt(0);
+      if (0xDC00 > lowSurrogateCode || lowSurrogateCode > 0xDFFF) {
+        throw this.createILLEGAL(char);
+      }
+      stream.readString(2);
```
disnet: was there supposed to be a `streamRead = true` here?

gabejohnson: You're right. I didn't catch your comment.

None of the current tests caught the issue. I'll add a few more tests and make the change.

PR #602, opened 2016-09-21
Initial work on refactoring modules and terms

disnet:
This is very much of work in progress but wanted to share my current status with refactoring our term and module handling. 

For context, `Term` has been our "class" for creating our internal AST. Our AST is basically the shift AST with some extra nodes to handle sweet specific syntax like `syntax` and `syntaxrec`. 

Our ways for manipulating a `Term` was an ugly combination of a visitor, reducer, and ad-hoc traversal methods. Also, there was no enforcement of how a `Term` could be constructed so we had lots of hard to track down bugs due to malformed `Term`s. 

So, the solution is to use more macros! The [sweet-spec](https://github.com/sweet-js/sweet-spec) repo defines all the `Term`s via the [sweet-spec-macro](https://github.com/sweet-js/sweet-spec-macro) using a syntax that is identical with flow's types:

``` js
declare class ClassExpression extends Expression {
  name?: BindingIdentifier;
  super?: Expression;
  elements: ClassElement[];
}
```

All these declarations are expanded into classes that check their arguments on construction. They also have a `reduce` method that runs a shift-style reducer to make traversal/manipulation nice and easy. 

Because the syntax is identical to flow we also can use the spec as a flow interface `.flow` file and get the benefits of type checking on our AST constructors. 

I'm pretty stoked about it. 

Anyway, everything is currently broken since I'm halfway through converting everything to use the new constructors but it should converge soon. 

/cc @gabejohnson @jlongster


jlongster:
I still haven't learned the structure of the new codebase, but pulling out a `sweet-spec` repo sounds great, especially if it can be fed into Flow!

src/reduce.js
```diff
@@ -0,0 +1,27 @@
...
+  } else if (List.isList(term)) {
+    return term.map(t => transform(reducer, t));
+  } else if (term instanceof Term) {
+    let state = new Term(term.type, {});
+    TermSpec.spec[term.type].fields.forEach(field => {
+      let v = transform(reducer, term[field]);
+      state[field] = v == null ? null : v;
+    });
+    if (typeof reducer[`reduce${term.type}`] !== 'function') {
+      throw new Error('Reducer failed to implemented handler for ${term.type}');
```
gabejohnson: This should be a template literal.

gabejohnson:
Looks much cleaner/easier to follow. Agree w/ jlongster about Flow.

src/reduce.js
disnet: True in general (template literals > strings) but this file is actually going to be deleted  :)

disnet:
Ok, so after a bunch of relatively straightforward work I’ve run into an issue I’d like folks opinions on (@gabejohnson I think we chatted about this before).

The key problem is that we have two distinct nestable structures: `Syntax` (nests delimiters) and `Terms` (the AST). This causes a bunch of complication in the code because you need to know which kind of object you are dealing with when you walk down the tree (e.g. when applying scopes). 

I think the right thing is to drop the idea of `Syntax` as a nestable structure. Rather than allow `Syntax` objects to have a `.inner` method, all `Syntax` objects are flat. `Term` objects are the only kinds of structures that are allowed to be nested and we introduce two new `Term` subclasses that represent a flat token and a delimiter token respectively.

Ok, so revised data types:

* `Token` — unchanged, basically what the tokenizer provides
* `Syntax` — almost the same but now does not have an `.inner()` method
* `SyntaxTerm` new `Term` node, empty just for inheritance purposes
* `RawDelimiter` a `Term` node extending `SyntaxTerm` to hold a list of `SyntaxTerm` nodes
* `RawSyntax` a `Term` node extending `SyntaxTerm` to hold a `Syntax` object

```js
class Token {
  type: string;
  value: any;
  // ...
}
class Syntax {
  context: Context;
  token: Token;
  addScope() {}
  val() { }
  // ...
}

class Term {}
// ...
class SyntaxTerm extends Term {}
class RawDelimiter extends SyntaxTerm {
  kind: 'parens' | 'square' | 'curly';
  inner: SyntaxTerm[];
}
class RawSyntax extends SyntaxTerm {
  value: Syntax;
}
```

Now the only thing a `Syntax` object needs to be concerned with is maintaining the lexical context rather than servings as *another* tree. Walking the entire tree can now be done very declaratively with a `CloneReducer`.

The new names are iffy but naming is hard. Thoughts? Does this seem reasonable?

gabejohnson:
So under this scheme `read` would return `SyntaxTerm[]`/`List<SyntaxTerm>`? Do `RawDelimiter`s throw out their opening and closing delimiter `Token`s (is this just folded into `kind`)?

gabejohnson:
What would be wrong with the `read` returning `SyntaxTree = List<Syntax|SyntaxTree>`. During macro expansion, getting a `List` back from a call to `next` would tell you that it's a delimiter. The `kind` of delimiter is known from the first and last elements in the `List`.

Or you could keep the current API and just handle `List -> inner` in `SyntaxOrTermWrapper`.

disnet:
> What would be wrong with the read returning `SyntaxTree = List<Syntax|SyntaxTree>`.

I like that. `read` doesn’t need to know about whatever higher level types the expander might need to use.

I think it’s still valuable to wrap a `SyntaxTree` in a `Term` inside the expander. This allows adding/removing scopes to be as simple as:

```js
let scope = // ...
t.reduce(class extends CloneReducer {
  reduceRawSyntax(term) {
    return new RawSyntax({
      value: term.value.addScope(scope)
    });
  }
})
```

Inside of a syntax transformer I think we need to keep its current API (`.inner` returns a `MacroContext`) rather than having `next` return a list since we want to provide `.expand` on syntax inside a consumed delimiter.

disnet:
@gabejohnson after rebasing on top of your readtable work I've run into a bug I'm having trouble tracking down.

If you checkout this branch you can see the problem by running `ava test/test-run-test262.js`. This runs Sweet over all the files in the test262 parser tests. It hangs after about 100 tests pass. Maybe we have a memory leak in the readtable somewhere? I'll keep digging but do you have any insight?

gabejohnson:
I suspect there's a case I didn't account for. I'm guessing the dispatch is happening but `readString` isn't being called for whatever reason. Then the same dispatch happens again. 

Do those tests include testing the contents of script tags? I intentionally didn't implement parsing HTML comments.

gabejohnson:
I won't be able to look until late tonight, at the earliest.

disnet:
Nothing with script tags. It looks like unicode troubles. See `1111.script.js` and `1252.script.js` in particular.

disnet:
Ah, actually there are some tests that are failing because of html comments. That's fine, we can address that whenever.

Also some tests are failing because of hex escape sequences in strings (e.g. `"\x20"`).

Looks like only 1111 and 1250 are causing an infinite loop.

gabejohnson:
I fixed what I could of the tests you added to the top of the exclusion list. I didn't touch HTML comments. There are a couple of bugs w/ `/` but I don't know whether it's a bug in `readDiv` in the default readtable, passing the `b`/`exprAllowed` flag around, something to do w/ line numbers or a problem w/ `isRegexPrefix`.

disnet:
Ok I think this is ready to merge.

Since we switched to `SyntaxTerm` this includes a breaking change:

```js
syntax m = ctx => {
  // previously since 1 was a Syntax
  #`1`.get(0).val() === 1;
  // now since 1 is a RawSyntax
  #`1`.get(0).value.val() === 1;
}
```

More ceremony but more consistent: everything the macro author interacts with is a subtype of `Term`. 

This change means we don't need the `SyntaxOrTermWrapper` anymore but I'll open a separate PR to replace it.

I'm going to hold off on publishing the major bump until I can get the wrapper and a few more things merged.

gabejohnson:
Really liking `module-visitor` and `sweet-loader`! The module caching should help a lot for large projects!

PR #611, opened 2016-12-10
Fix parse errors

gabejohnson:
While attempting to replicate the performance test https://github.com/shapesecurity/shift-parser-js/blob/ce88859a4636ba1d0d06d7d3e0d89ade767c8233/profile.js I discovered a bug with the enforestation of `this` in new expressions.


disnet:
Nice!

PR #612, opened 2016-12-19
Fix string and identifier reader

gabejohnson:
Fixed the reader bugs exposed by the test262 test suite. I took the liberty of checking the rest of the tests that were excluded and removed those that now pass.

Aside from the HTML comment tests there are a couple that are failing b/c of misreading `/`. I'm not sure if I missed something in my implementation, or this was already a problem.


disnet:
Nice, thanks!

Yeah don't worry about `/`. I did the initial implementation for ES5 and still haven't gotten around to thinking through the ES2015 cases so that's probably what's going on there.

Issue #615, opened 2016-12-27
Too many syntax object types in macros


disnet:
So it turns out we can't quite remove `SyntaxOrTermWrapper` like I thought. I'd forgotten about the `inner` method that returns a new `MacroContext`, which is necessary because it allows `expand` inside of delimiters.

So currently a macro can get:
* a `SyntaxOrTermWrapper` via `ctx.next().value`
* a `SyntaxTerm` via `#``foo``.get(0)` 

Previously it was:
* a `SyntaxOrTermWrapper` via `ctx.next().value`
* a `Syntax` via `#``foo``.get(0)` 

I'm not a fan of macro authors needing to know about two different representations of the syntax, but I'm not really sure what else to do about it. Maybe the distinction can be papered over by a higher-level API macro authors import in? Maybe we modify `SyntaxOrTermWrapper` to behave a bit more like  a `Term` but with the `inner` method still being special?

gabejohnson:
What about exposing a `ctx.createContext` which takes a `RawDelimiter`?
```js
let syntax m = ctx => {
  const inner = ctx.next().value.val();
  const innerCtx = ctx.createContext(inner);
// ...
}
m(1,2,3)
```

disnet:
Yeah, that might work. I think we had something like that early on but switched to `inner` on the wrapper for a more ergonomic API. Consistency in the types is more important than a slightly more convenient way to "contexify" a delimiter.

gabejohnson:
That was my thinking. With your changes wrapping all of the `SyntaxTerms` is just unnecessary overhead.

Not sure about the name though. `wrap`? `create`? `contextify`?

PR #617, opened 2016-12-31
Drop wrapper in macro context

disnet:
Addresses #615.

Dropping `SyntaxOrTermWrapper` and replacing its `.inner` method with a `contextify` (strawman name but I kind of like it) method on `MacroContext`.

This is doubling down on the breaking change of everything being a `Term`.


gabejohnson:
`contextify` is what Node.js calls it when a VM context is created. So there's precedent.

PR #619, opened 2017-01-05
Add ability to select extensions to resolve

disnet:
By default `resolve` only looks for files with `.js` as an extension. This adds the ability for the `NodeLoader` to enable additional extensions to resolve.

### DIFF
--- a/src/node-loader.js
+++ b/src/node-loader.js
@@ -7,9 +7,11 @@ import vm from 'vm';
 import Store from './store';
 
 export default class NodeLoader extends SweetLoader {
+  extensions: ?string[];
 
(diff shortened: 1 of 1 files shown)


gabejohnson:
Have you thought about this wrt `lang`? I was thinking about a file that maps file extension/paths/globs -> language. This would obviate the need for `lang` directives in every file w/ a particular extension.

disnet:
I'm really not a fan of semantic file extensions. Much better to be declarative in your code imo.

That said, I wouldn't be opposed to adding the ability to do this. I would just strongly discourage people from using it 😄

PR #623, opened 2017-01-07
Add partial call expression term

disnet:
This fixes parsing of code like `s => s.m('foo')`. Before merging I need to find the disabled parser test that covers this.

@gabejohnson this should fix the problem you were running into with loading `reader/utils.js`.


gabejohnson:
That did the trick!

PR #626, opened 2017-01-09
Reader modes

gabejohnson:
Adds support for reader modes
```js
const noYs = getCurrentReadtable().extend({
  key: 'y',
  mode: 'terminating', // 'y' will now terminate Identifier parsing effectively acting as a delimiter
  action(stream) {
    throw Error('y is an invalid character');
  }
});

const useZs = getCurrentReadtable().extend({
  key: 'z',
  mode: 'non-terminating', // 'z' can still be used in Identifiers
  action: // ...
});

const colonese = getCurrentReadtable().extend({
  key: ':',
  mode: 'dispatch', // '#:' will dispatch to action and acting as 'non-terminating'
  action: // ...
});

// delegate mode in progress
const twoWaysToAdd = getCurrentReadtable().extend({
  key: 'Σ',
  mode: '+' // now Σ will return a '+' token,
  delegateTable: getCurrentReadtable()
});
```

I'm currently working on the delegate mode. I'm not sure how useful it's going to be and it does complicate things a bit.

I'm only including it b/c Racket has it. If I can't get it in a day or two I might drop it.


gabejohnson:
@disnet I think this is ready to go. I ended up scraping the delegation. I'm not exactly sure how it's used in Racket, but it causes a strange overloading of `make-readtable` and `readtable-mapping` https://docs.racket-lang.org/reference/readtables.html

Calling `make-readtable` with `char, like-char, readtable` dispatches to the `like-char` mapping in `readtable`.

Calling getting `char, #f, #f` back from `readtable-mapping` doesn't appear to give access to the readtable that was passed to `make-readtable`. Thus you can't (necessarily) get access to the mapping that was delegated to.

Do you have any strong feelings either way?

disnet:
This is looking really great! No I don't have any strong feelings about delegation. I we run into a need for something like it down the road we can always revisit.

elibarzilay:
FWIW, I don't think that the delegation thing was really useful.  IIRC
(and that's a big IF), the real thing to look at are readtables in
Common Lisp -- where you can get the binding of a specific character
from an existing readtable and then use it to extend it using a
different character or use it with a different readtable.  I think that
my impression at the time was that Racket didn't expose handlers
"enough" -- you get a `#f` for the default reader handlers, so
delegation is more conveniet in those cases.

disnet:
@elibarzilay thanks for the context!

gabejohnson:
Yeah thanks @elibarzilay. I just looked over the CL readtable API and I think we have most of it covered. Except dispatch macros. If I'm reading this correctly, in CL you can define multiple dispatch characters (not aliases). So if `#` and `@` are dispatch macro chars, `#!` and `@!` can be mapped to different procedures.

@disnet I think we should consider extending the dispatch capabilities before exposing readtables.

Currently we have:
```javascript
// # is the default dispatch macro character
// we can alias it like so
const currentTable = getCurrentReadtable();

setCurrentReadtable(currentTable.extend({
  key: '@',
  mode: 'non-terminating',
  action: currentTable.getEntry('#').action
}));
``` 
and to add a dispatch macro we do
```javascript
setCurrentReadtable(currentReadtable.extend({
  key: "'",
  mode: 'dispatch',
  action : function(stream) { ... }
}));

// #' and @' will do the same thing
```
We could change this to
```javascript
currentReadtable.extend({
  key: '@',
  mode: 'non-terminating',
  readtable: someDispatchReadtable // the key 'readtable' tells extend that we're defining a dispatch char
}, {
  key: "'",
  dispatchKey: '@',
  action: function (stream) { ... }
});
```
Omitting the `readtable` key could cause `extend` to just make a new readtable.

elibarzilay:
@gabejohnson, IIRC (which again, is not much, I dealt with that stuff more than a decade ago), losing them is not a big deal.  I just happens that `#` is used for a bunch of things, so it's a kind of a short way to be able to define `#_` sequences independently.  It's fine to just have basic functionality and let people do that themselves when/if needed.  Alternatively, you can allow the mapping of a character to itself be a readtable...

(Another side-comment: Emacs and interactive functionality had some noticeable influence on Lisp in general, and you can see how going down that path quickly gets you to areas close to Emacs.)

PR #618, opened 2017-01-04
Add support for more import/export forms

disnet:
This PR does some more refactoring of the module handling code and adds support for:

- default imports/exports
- export specifiers with renaming (e.g. `var x = 1; export { x as y }`)
- import namespaces (`import * as T`)

PR #629, opened 2017-01-20
Remove module tests from the passExclude list

gabejohnson:
Since #618 & #624 several of the conformance tests for modules are now passing. This PR removes the passing tests from the exclusion list.


disnet:
Thanks!

PR #635, opened 2017-01-26
Simplify prefix so only TokenTree is used

disnet:
The regex prefix stuff doesn't actually need to know anything inside of delimiters (only that it is a paren or bracket or brace). So, rather than force `isParens` etc know about terms, unwrap delimiters when putting them in the prefix.

/cc @gabejohnson

### DIFF
--- a/src/reader/token-reader.js
+++ b/src/reader/token-reader.js
@@ -5,7 +5,6 @@ import defaultReadtable from './default-readtable';
 import { List } from 'immutable';
 import CharStream, { isEOS } from './char-stream';
 import { EmptyToken } from '../tokens';
-import Syntax from '../syntax';
 
 import type { StartLocation, Slice } from '../tokens';
 
@@ -71,7 +70,7 @@ class TokenReader extends Reader {
     : this.createError('Unexpected end of input');
   }
 
-  readToken(stream: CharStream, ...rest: Array<any>): Syntax {
+  readToken(stream: CharStream, ...rest: Array<any>) {
     const startLocation = Object.assign({}, this.locationInfo, stream.sourceInfo);
     const result = super.read(stream, ...rest);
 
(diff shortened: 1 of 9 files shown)


src/reader/token-reader.js
```diff
@@ -108,6 +108,13 @@ class TokenReader extends Reader {
   }
 }
 
+function unwrapToken(tok: List<T.SyntaxTerm>) {
+  if (List.isList(tok)) {
+    return List.of(tok.first().value, tok.last().value);
```
gabejohnson: Reader macros won't be able to examine the contents of the delimiters.

gabejohnson:
So the change is fine if the only thing we're reading is JS. But since we decoupled the tokenization we aren't necessarily reading JS, we're writing JS (objvs). But what if I'm translating Python into JS? Then I have to encode significant whitespace as braces. In that case Python's scoping rules are probably sane enough that I don't have to worry about drilling down into `TokenTree`s.

But my point is that I can't predict what language design someone is going to come up with. And I think that we shouldn't destroy any information someone may need if we don't have to.

Ideas:

1. Do two passes

Pro: simple
Con: another traversal

2. Unify `Token`s and `Term`s

Pro: eliminates special cases
Con: `Token`s aren't `Term`s (but then again neither are `RawSyntaxTerm`s)

3. Return a lazy iterator from the reader. Similar to what we're doing in the enforester. Then the second pass isn't an entire second traversal.

Pro: we could be lazy all the way through the system and (best case) do a single traversal kicked off by `SweetToShiftReducer`. Until codegen that is.
Con: lots of work

4. Don't listen to me. I don't know what I'm talking about

Pro: no work
Con: maybe I am right and we have to fix for v5

disnet:
Options 2 and 3 are really interesting. Let me think about this some more but I suspect the right choice at the moment is to go with 1 for now and then pick either 2 or 3 later when we start focusing on perf.

gabejohnson:
I find 3 particularly attractive and thinking about it more it really shouldn't be that much work. But I agree we should wait until we're close to feature complete and working on perf.

PR #637, opened 2017-01-28
Replace readtable, reader and char-stream w/ readtable npm package

gabejohnson:
https://github.com/gabejohnson/readtable


disnet:
:+1:

PR #638, opened 2017-01-30
Unify Terms and syntax

disnet:
Still work in progress. At the moment we've just got the readtable returning a TokenTree and then wrapping in both a term and a syntax. Next up is dropping `Syntax` and moving the context onto `Term`.


test/test-reader.js
```diff
@@ -1,292 +1,248 @@
...
+  t.true(T.isIdentifier(iinner, 'bar'));
 });
 
-test('should read a regex when it follows a keyword', function () {
-  let r = read('return /42/i');
-
-  expect(r.get(1).isRegularExpression()).to.be(true);
-  expect(r.get(1).val()).to.be('/42/i');
+test.skip('should handle escaped string templates literals inside a syntax literal', t => {
+  let [r] = read('#`x = \`foo\``');
```
gabejohnson: Should this be ```let [r] = read('#`x = \\`foo\\``');```? 

```'`x = \`foo\``'``` shouldn't parse properly. Should it?

File: test/test-reader.js [hunk 2]
```diff
@@ -1,292 +1,248 @@
...
+  t.true(T.isSyntaxTemplate(r));
+  let [open, x, eq, str, close] = r;
+  t.true(T.isIdentifier(x, 'x'));
+  t.true(T.isPunctuator(eq, '='));
+  t.true(T.isString(str, 'foo'));
```
gabejohnson: This should be a check to see that `str` is a template.

test/test-reader.js
disnet: Ah, I think you are right

disnet:
closed in favor of #639.

Issue #600, opened 2016-09-09
Missing statement in output


jlongster:
A funny edge case:

``` js
if(true) {
  foo.bar("f")
  x = y
}
```

The output of this file is:

``` js
if (true) {
  x = y;
}
```

The `foo.bar` is gone. It seems to only happen with a member lookup. Unfortunately for me, it was a `console.log` that was missing so it was quite confusing....

Actually, looks like this repros it:

``` js
if(true) {
  foo.bar
  x = y
}
```

jlongster:
Oh, weird, all these edge cases repro:

```js
x
foo.bar
x = y
```

```js
foo.bar
x = y
```

```js
x
foo.bar()
x = y
```

```js
x
foo.bar("hi")
x = y
```

gabejohnson:
Fixed in https://github.com/sweet-js/sweet.js/pull/641

PR #641, opened 2017-02-21
Fix IdentiferExpression encountered after New/CallExpression

gabejohnson:
```javascript
{
  foo = bar()
  baz = foo
}
// and
{
  foo = new bar()
  baz = foo
}
```
both result in:
```javascript
{
  foo = baz = foo;
}
```

This is because after enforestation of either a `CallExpression` or a `NewExpression` the while loop in `enforestLeftHandSideExpression` was either entered (in the case of `NewExpression`) or continued (in the case of `CallExpression`).

This PR fixes that by breaking or returning respectively.

### DIFF
--- a/test/assertions.js
+++ b/test/assertions.js
@@ -1,17 +1,41 @@
-import read from '../src/reader/token-reader';
-import { compile } from '../src/sweet';
-import { Enforester } from '../src/enforester';
 import { List } from 'immutable';
+
+import read from '../src/reader/token-reader';
+import { compile, parse } from '../src/sweet';
+import { Enforester } from '../src/enforester';
 import StoreLoader from '../src/store-loader';
 
 export const stmt = x => x.items[0];
 export const expr = x => stmt(x).expression;
 export const items = x => x.items;
 
-
 export function makeEnforester(code) {
   let stxl = read(code);
   return new Enforester(stxl, List(), {});
+}
+
+function getAst(code) {
+  const store = new Map();
+  store.set('main.js', code);
+
+  const loader = new StoreLoader(__dirname, store);
+  return parse('main.js', loader);
+}
+
+function testParseWithOpts(t, acc, code, expectedAst) {
+  try {
+    t.deepEqual(expectedAst, acc(getAst(code)));
+  } catch (e) {
+    throw new Error(e.message);
+  }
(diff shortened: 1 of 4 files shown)


gabejohnson:
I'd like to add a test for this.

gabejohnson:
@disnet I just noticed that the `test262` tests don't check to see that the parse was correct. Just that it returned a result. How would you suggest I test this? We got rid of all of the parser tests when you added the `test262` stuff.

gabejohnson:
This is not the answer as this:
```javascript
{
  foo = new bar()['a']
  baz = foo
}
```
results in:
```javascript
{
  foo = baz = foo
}
```
It appears to actually be what to do when you encounter an identifier in the loop. I think breaking if `this.term != null` is the way to go.

gabejohnson:
Fixes #600

disnet:
Nice job tracking this down!

For tests I think what you should do is assert that the AST for the code with a semicolon is the same as the AST without the semicolon.

gabejohnson:
@disnet can you review this when you get a chance?

test/test-parse.js
```diff
@@ -0,0 +1,9 @@
+import test from 'ava';
+
+import { items, testParseComparison } from './assertions';
+
+test('CallExpression followed by identifier', testParseComparison, items,
+     'x\nfoo.bar(1,2)\nx', 'x;\nfoo.bar(1,2);\nx;');
```
disnet: use template literals with an actual newline to make the tests easier to follow

test/assertions.js
```diff
@@ -1,19 +1,64 @@
...
+    throw new Error(e.message);
+  }
+}
+
+export function testParse(t, acc, code, expectedAst) {
+  return testParseWithOpts(acc, code, expectedAst);
+}
+
+export function testParseComparison(t, acc, codeA, codeB) {
+  testParse(t, acc, codeA, acc(getAst(codeB)));
```
disnet: I think you should just use `t.deepEqual` for this kind of tests and avoid the custom "object equality modulo `<<hygiene>>` string". The tests you added don't require the custom stuff and I would prefer if we could avoid reintroducing tests that assert on hand written ASTs.

PR #643, opened 2017-02-26
Add experimental infix macro support

disnet:
Experiment is experimental.

Macros now take a second argument which is a single element list with the previous term.

In the old pre-1.0 days we allowed a list of *all* the previously enforested terms but this didn't behave well so I'm starting out with just a single term. We can argue our way into relaxing this restriction later.

Note that infix macros are contained within implicit delimiters so:

```js
syntax m = (ctx, prev) => {
  return prev;
}
1 + m
```

throws an error since `prev` is the empty list.

### DIFF
--- a/src/enforester.js
+++ b/src/enforester.js
@@ -137,6 +137,7 @@ export class Enforester {
     if (this.rest.size === 0) {
       this.done = true;
     }
+    this.prev = List.of(result);
     return result;
   }
 
@@ -1883,7 +1884,7 @@ export class Enforester {
 
       let ctx = new MacroContext(this, name, this.context, useSiteScope, introducedScope);
 
-      let result = sanitizeReplacementValues(syntaxTransform.value.call(null, ctx));
+      let result = sanitizeReplacementValues(syntaxTransform.value.call(null, ctx, this.prev));
       if (!List.isList(result)) {
         throw this.createError(name, 'macro must return a list but got: ' + result);
       }
--- a/src/token-expander.js
+++ b/src/token-expander.js
@@ -91,8 +91,7 @@ export default class TokenExpander extends ASTDispatcher {
     if (stxl.size === 0) {
       return List(result);
     }
-    let prev = List();
-    let enf = new Enforester(stxl, prev, this.context);
+    let enf = new Enforester(stxl, List(), this.context);
 
     while (!enf.done) {
       result.push(this.dispatch(enf.enforest()));
--- a/test/test-macro-expansion.js
+++ b/test/test-macro-expansion.js
@@ -241,3 +241,10 @@ syntax m = ctx => {
 }
 output = m
 `, 26);
+
+test('should support infix macros', evalWithOutput, `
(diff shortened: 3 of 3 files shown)


gabejohnson:
What were the issues with access to all previous terms?

disnet:
I seem to remember it being related to concerns about consuming already defined bindings and splitting terms (though that might be solved by just keeping it contained to implicit delimiters). Maybe @natefaubion can remember more?

btw, just realized this PR is wrong. Need to put it on the macro context and have it be consumed.

gabejohnson:
In the above example, the implicit delimiter is created as a result of the ongoing `BinaryExpression` enforestation?

natefaubion:
The implications for hygiene were probably the biggest concern. Otherwise it was extremely awkward for pattern matching (you couldn't use custom pattern classes), because it was actually matched in reverse. When combined with custom operators, the implementation and edge cases where very opaque. Unless there is a really compelling use case, I'd advise not to go down that route again 😜. The use case at the time was arrow functions, which is moot now, but could also have been solved with single term look behind.

disnet:
> In the above example, the implicit delimiter is created as a result of the ongoing BinaryExpression enforestation?

yep

> The implications for hygiene were probably the biggest concern.

Ah! You made me think of an example:

```js
let x = 1;
syntax m = ctx => {
  let decl = ctx.prev();
  return #`function foo() { ${decl} }`;
}
function f() {
  let x = 2; // this gets consumed by m
  m
  return x;  // unbound? bound to the outer let x?
}
```

This happens even if we only allow a single term look behind. Totally unclear what should happen if a macro moves an already registered binding into a different scope.

gabejohnson:
> Totally unclear what should happen if a macro moves an already registered binding into a different scope.

We could just disallow this. In fact, I think we should only return expressions from `prev`.

gabejohnson:
Also, do you want `prev` to be a getter? Or keep it as a method in case we change our minds about single term lookbehind?

disnet:
> We could just disallow this. In fact, I think we should only return expressions from prev.

Is `{ var x }` an expression? No, but the enforester (where the infix macro expansion would happen) doesn't know that because processing the insides of a delimiter are deferred until later.

There are lots of ad-hoc restrictions we could impose but I'd much prefer a principled approach because of gotchas like that. Not sure if there is one though.

I think we should put this on ice for now. We can come back to it if we figure out "the right way" or we have a sufficient use-case covered by an ad-hoc restriction.

gabejohnson:
> processing the insides of a delimiter are deferred until later.

You mean in the expander?

It's my opinion that if you're making an infix macro you're doing something akin to _complicated_ syntactic surgery. You have to be very careful to expand the previous term and inspect it before you go tossing it in into a different scope. The very example you give should be put in the docs to illustrate how tricky infix macros are.

gabejohnson:
We'll have to remember to reset `prev` when `reset` is called.

disnet:
On ice for now. Can reevaluate later

PR #644, opened 2017-03-11
Add helpers

disnet:
This adds a helper module for macro authors. This will make it easier to change our underlying data model without causing pain for macro authors who use the new helper library.

It should replace the various methods we had on `Syntax` objects. Use looks something like:

```js
import * as H from 'sweet.js/helpers' for syntax;

syntax m = ctx => {
  let v = ctx.next().value;
  if (H.isIdentifier(v, 'foo') {
    return H.fromString(v, 'bar');
  }
  return H.fromString(v, 'baz');
}
m foo; // expands to 'bar'
```


helpers.js
```diff
@@ -0,0 +1,124 @@
...
+}
+
+export function isIdentifier(obj, val) {
+  return check(obj, TypeCodes.Identifier, val);
+}
+
+export function fromIdentifier(obj, x) {
+  return obj.value.fromIdentifier(x);
+}
+
```
gabejohnson: I'm wondering about the order of parameters here. It might be nice to swap them so I can:

```js
const isFor = H.isKeyword.bind('for');
const makeFor = H.fromKeyword.bind('for');
```

gabejohnson: Thinking on it further, the argument could be made that the scope inside the macro is the most stable argument. So I don't know.

disnet: Good observation. The `is*` functions should definitely be swapped so to keep things consistent the `from*` need to be swapped too; I agree the `from*` could be argued either way but consistency should win out.

disnet:
> The only thing that bugs me is the TypeCodes kludge. When we figure out what to do w/ Syntax I think we should write a token spec and have helpers and internals both depend on it.

Agreed. I hate the kudge too.

> On a side note I see you've hopped on the prettier train. I hadn't looked at it closely before. It really opens things up!

Yeah I love it! I'll hook it up to our build process soon.

helpers.js
disnet: Oh! I'm silly, we can't swap them because often you don't care about the value. You just want to know if `isIdentifier(x)`.

gabejohnson: Ahh..right. We could add `equalsIdentifier(x, obj)` and restrict `is*` to be a single argument predicate. This is explicitly comparing two things vs overloading a predicate. But that would increase the API surface a bit.

disnet: Yeah I don't want to increase the API surface.

But you're right the overloading of `is*` is problematic. How about doing what you suggested and restrict `is*` to a single argument and then let the equality stuff be done by the consumer:

```js
const isFor = x => isKeyword(x) && unwrap(x).value === 'for';
const equalsIdentifier = (x, obj) => isIdentifier(obj) && unwrap(obj).value === x
```

gabejohnson: Perfect!

disnet:
I've rearranged the tests a little because originally the `test-handlers.js` file I added wasn't actually being run (forgot to update the files list). Now new tests files will just be picked up automatically.

Let me know if everything looks good.

gabejohnson:
These won't work w/ `MacroContext#name`. It's a `Syntax` object, not `RawSyntax`. We can either pass the `RawSyntax` to the `MacroContext` constructor or wrap it again on the call to `name`.

PR #650, opened 2017-03-18
Throw error when missing commas in Array and Object expressions

### LINKED ISSUE #605: Extraneous comas in modified example
Hello,

I just modified a simple example and i have an invalid result:

``` javascript

syntax m = ctx => {
  let delim = ctx.next().value;
  var arr = [];
  for (let item of delim.inner()) {
    arr.push(item);
    arr.push(item);
  }
  return #`[${arr}]`;
}

m {
  1, 2, 3, "sdf"
}
```

output:

`[1, 1, , 2, 2, , 3, 3, , "sdf", "sdf"]`

gabejohnson:
Fixes #605

Formerly:
```js
[1 2 3]; // [1, 2, 3];

let [a b c] = []; // let [a, b, c] = [];

({ a: 1 b: 2 c: 3 }); // ({a: 1, b:2, c: 3});

let { a b c } = {}; // let { a, b, c } = {};

let [...rest, a] = []; // let [...rest_0] = [];

[...[],]; // [...[],,];
```
All of these cases save the last should throw (it was simply parsing incorrectly). Now they do:
```js
[1 2 3]
// Error: unexpected token
// __2__ 3

let [a b c] = [];
// Error: unexpected token
// __b__ c

({ a: 1 b: 2 c: 3 })
// Error: unexpected token
// __b__ : 2 c : 3

let { a b c } = {};
// Error: unexpected token
// __b__ c

let [...rest, a] = [];
// Error: Rest element must be last element in array
// , a

[...[],]; // [...[]];
```


gabejohnson:
@disnet I realized after the fact that I pushed this branch to the main repo. Do you want me working in here or should I delete the branch and resubmit from my repo?

gabejohnson:
This isn't quite right yet. I messed up on the with the spread operator.

gabejohnson:
It's ready to review.

disnet:
This looks great! Are there some test262 tests we can enable now to catch this?

disnet:
btw creating branches on this repo is totally fine

gabejohnson:
> Are there some test262 tests we can enable now to catch this?

If not, we should add some.

We need to start running the `fail` and `early` tests. I'm sure that would give us _tons_ of work to do 😄

gabejohnson:
Do you want me to submit a PR to enable the `early` and `fail` tests before merging this one? It would be nice to make sure we have a regression test in place.

disnet:
Yeah that would be great

gabejohnson:
These aren't bugs w/ the new code, I think the problem is [here](https://github.com/sweet-js/sweet.js/blob/e55c29f0c43c71d68dd784e35cb7a7b6b2142965/src/enforester.js#L1387).  We should be a checking to see if `=` follows.

gabejohnson:
While ☝️ is true, the problem is greater. The [`transformDestructuring`](https://github.com/sweet-js/sweet.js/blob/a93eac62934ac62c7c091e0b1ca1b42d522b847a/src/enforester.js#L1403) trick won't work as the rules for what can follow a rest element are different than those for a spread element.

This is going to bite us again once object rest/spread lands.

I propose a refactor that uses double lookahead for in instances where there might be an `AssignmentExpression` and deals with the bindings right away.

As far as this PR goes, do you want to hold on it? Or merge it, comment out the offending tests and proceed with a refactor?

disnet:
Let's merge it now and then refactor.

PR #652, opened 2017-03-22
Add import/export transforms

disnet:
Allows macros to bind on the import and export keywords. Previously this was an error:

```js
syntax import = ctx => ...
```


gabejohnson:
Should we be doing this in more places where we test for a specific keyword? [`enforestExportDeclaration`](https://github.com/sweet-js/sweet.js/blob/a08cbde074acec14dc1f20433e4f66b77c28ad3b/src/enforester.js#L191) for instance.

If `default` is bound to a macro, the transform isn't going to take place. There are a couple of places in `enforestLeftHandSideExpression` where this could be an issue as well.

disnet:
Good call. I've updated the PR, let me know any place else I've missed.

gabejohnson:
https://github.com/sweet-js/sweet.js/blob/5843ed5f8a1b6a62c09ce0e5dfd3b10289e2da74/src/enforester.js#L452

https://github.com/sweet-js/sweet.js/blob/5843ed5f8a1b6a62c09ce0e5dfd3b10289e2da74/src/enforester.js#L1214

https://github.com/sweet-js/sweet.js/blob/5843ed5f8a1b6a62c09ce0e5dfd3b10289e2da74/src/enforester.js#L1214

https://github.com/sweet-js/sweet.js/blob/5843ed5f8a1b6a62c09ce0e5dfd3b10289e2da74/src/enforester.js#L1245

But then we have the question: how atomic is an enforestation?

https://github.com/sweet-js/sweet.js/blob/5843ed5f8a1b6a62c09ce0e5dfd3b10289e2da74/src/enforester.js#L383

This is deep in the enforestation of a `SwitchStatement`.

disnet:
Thanks for catching those. I think the ones in `enforestAssignmentExpression` and `enforestStatement` were technically "fine" because macros in the position are completely expanded before reaching the check, however doing the transform check is more correct anyway.

The one deep in the switch statement was a bug, if `class` was bound to a macro it needed to fall through to `enforestStatement` which triggers macro expansion.

gabejohnson:
Found a few more:
https://github.com/sweet-js/sweet.js/blob/678709944272837750117b86d1fe53127c36c6b0/src/enforester.js#L453
https://github.com/sweet-js/sweet.js/blob/678709944272837750117b86d1fe53127c36c6b0/src/enforester.js#L1246
https://github.com/sweet-js/sweet.js/blob/678709944272837750117b86d1fe53127c36c6b0/src/enforester.js#L1347

These would be purely for correctness though.

PR #654, opened 2017-03-23
Add test262 failure tests

gabejohnson:
I've added the test262 failure tests and commented out the ones that fail 😛 Most of the failures are `*.script.js` tests, so it looks worse than it is.

I started to add the `early` tests as well, but _most_ of them were failing and I decided to stop adding exclusions. I'm leaving the exclusions I _did_ add for now. They'll have to wait until we tackle early errors.

We should think about running the `pass` module tests and opening issues for the failure cases.


disnet:
> I started to add the early tests as well, but most of them were failing and I decided to stop adding exclusions.

I think these are very low priority for us. We need to be fairly loose during expansion especially if we introduce some sort of infix form.

PR #660, opened 2017-03-30
Change webpack output from AMD to UMD to support global.sweet

gabejohnson:
In the Gitter room we got a question about using sweet in the browser. The editor already uses a webpack bundle 😄. But it's AMD ☹️.

I changed the output type to `umd` to support all of the things. Now it should work with require.js, as a global and in node.


gabejohnson:
We should consider having a `sweet-cli` and a `sweet-compiler` package.

PR #664, opened 2017-04-02
Switch to string directive

disnet:
Switch to using the `'lang sweet.js'` instead of `#lang "sweet.js"` directive.

### DIFF
--- a/src/sweet-loader.js
+++ b/src/sweet-loader.js
@@ -26,16 +26,23 @@ export type Context = {
   store: Store,
 };
 
+export type LoaderOptions = {
+  noBabel?: boolean,
+  logging?: boolean,
+};
+
 export default class SweetLoader {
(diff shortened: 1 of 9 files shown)


src/sweet-loader.js
```diff
@@ -164,7 +168,11 @@ export default class SweetLoader {
     return new Store({});
   }
 
-  compileSource(source: string) {
+  compileSource(source: string, metadata: {}) {
+    let directive = getLangDirective(source);
+    if (directive == null && metadata.enforceLangPragma) {
+      return new SweetModule(List.of());
+    }
```
disnet: It should definitely not throw.

The idea is to avoid unnecessary expansions of runtime imports since in the vast majority of cases your imports are to non-sweet code.

Logging is reasonable though, I'll add that.

disnet:
Logging added.

PR #665, opened 2017-04-03
Implement custom operators

### LINKED ISSUE #517: Add back custom operators
Pre 1.0 had the ability to define custom operators with precedence and associativity. It would be nice to have that back. Some syntax ideas:

``` js
operator ** 14 right = function (left, right) {
  return #`Math.pow(${left}, ${right})`;
}
```

Most of the plumbing for this features is already done. Just need to decide on the right syntax.

disnet:
Fixes #517.

Easier than I thought it would be.

I chose to go with using the same declarator form that `syntax` and `syntaxrec` use. Made the implementation much simpler than other forms since it can just piggyback on the work done for `syntax`/`syntaxrec` in terms of binding and loading into the compiletime env.

It's not the most user-friendly syntactic form but we can always provide a macro the give some sugar (I suspect a declaration form would work nicely but we can play with that later).

I've just implemented binary and prefix unary. Postfix unary requires a few more changes that I'd like to do later.

```js
operator >>= left 3 = (left, right) => {
  return #`${left}.chain(${right})`;
}
class IdMonad {
  constructor(value) {
    this.value = value;
  }
  chain(f) {
    return f(this.value);
  }
}

function Id(value) {
  return new IdMonad(value);
}

let result = Id(1) >>= v => Id(v + 1)
                   >>= v => Id(v * 10);
```


disnet:
> Postfix unary requires a few more changes that I'd like to do later.

Actually postfix was straightforward too so I'll include that in this PR too.

PR #671, opened 2017-04-06
Expose helpers in the editor

disnet:
This just uses require.js to load the helpers module as text and then stick the string in the loader store.

A more correct solution is to actually implement a browser loader that uses `window.fetch` but we first need to make Sweet play nice with an async loader.

I'm also including with this PR a commit to hide the editor controls we don't currently use.


gabejohnson:
Should we start ignoring browser build artifacts? Actually it kinda makes sense to move the site specific stuff to another repo.

disnet:
Yep, the editor should be extracted to its own repo. After the release :)

PR #681, opened 2017-04-14
Update prettier to 1.0

disnet:


test/parser/test-run-test262.js
```diff
@@ -437,7 +437,8 @@ const failExcluded = [
   '682.script.js',
 ];
 
-const earlyExcluded = [ // eslint-disable-line no-unused-vars
+const earlyExcluded = [
+  // eslint-disable-line no-unused-vars
```
gabejohnson: Prettier is in conflict w/ eslint comment directives? There's an issue for it https://github.com/prettier/prettier/issues/1267

Should we hold off and see if there's a fix first? Or do you want to use the workaround suggested in the issue?

disnet: `// eslint-disable-next-line` is the best solution I think, will go with that

Issue #632, opened 2017-01-21
Syntax and Terms - this isn't even my final form!


disnet:
Sweet internals geekery.

In #602 I tried to unify our syntax nesting structures by adding `RawSyntax` and `RawDelimiter` to the `Term` subclasses and changing `Syntax` to never wrap a delimiter.

This was good because we didn't really need two trees hanging around but the current state of things is still messy and led to bugs like #625.

I *think* the right solution is to just completely remove `Syntax`. The only real job it's doing anymore is holding on to the lexical context (scopesets) which can just as easily be done on `Term`s.

So the proposal is to:

* remove `Syntax`
* add `context` on all `Term` objects
* add `addScope` and `removeScope` on `Term`s
* add `from*` methods correctly on `Term`s
* change `Syntax::resolve` to be a standalone function that works over `IdentifierExpression` and `BindingIdentifier` terms

Seem reasonable?

gabejohnson:
Now you're speaking my language! I actually started to ask you about this on Gitter a couple of weeks ago but was too tired to get my thoughts straight.

It would simplify implementation of the default readtable in some places (though I'll have to make some changes to https://github.com/sweet-js/sweet.js/blob/master/src/reader/utils.js#L234-L241. And maybe a few others. But I'm all for it!

disnet:
> (though I'll have to make some changes to https://github.com/sweet-js/sweet.js/blob/master/src/reader/utils.js#L234-L241)

Funny story, I've already done that 😄Was yak shaving on it earlier this week, little more cleaning and I'll push it up for your review.

I actually think it might make sense to extract the reader+readtable to its own repo/package since it's generally useful beyond just Sweet.

gabejohnson:
> I actually think it might make sense to extract the reader+readtable to its own repo/package since it's generally useful beyond just Sweet.

I was holding off suggesting this. It's the reason why I implemented `Reader` separately from `TokenReader` in the first place.

`Reader`, `Readtable` and `CharStream` should all be extracted together.

Something that's still nagging me is how `TokenReader::locationInfo` is separate from `CharStream::sourceInfo`. I did this on purpose as line and column numbers don't have any "real" meaning in a character stream but it still feels icky.

I would also like to answer the questions I brought up in https://github.com/sweet-js/sweet.js/pull/626#issuecomment-274308895.

gabejohnson:
Thinking about it some more, getting rid of `Syntax` would allow me to:

[...]

PR #704, opened 2017-05-09
Update to sweet-spec with better flow support

disnet:


src/enforester.js
```diff
@@ -63,7 +63,7 @@ const EXPR_LOOP_OPERATOR = {};
 const EXPR_LOOP_NO_CHANGE = {};
 const EXPR_LOOP_EXPANSION = {};
 
-function getLineNumber(x: Syntax | T.Term) {
+function getLineNumber(x: any) {
```
gabejohnson: Why are we relaxing the type constraints for the arguments on all of these functions/methods?

disnet: because flow doesn't pass otherwise 😅 
 
My plan was to just get flow to pass and then refactor to better types in a separate PR but looking at the diff now it's probably better to do the refactoring now.

I'll add some more commits so we don't regress on the types.

gabejohnson: Sounds good. You might want to add `// @flow` to the top of all files using type annotations at the same time. Unless you want to save that for another PR.

disnet: On second thought, unless you have any objections I'll do the refactoring in a separate PR. I realized while trying to put the types back in that the right way to get flow to pass essentially requires syntax/term unification #632. I'll work on that next but probably worth separating it out.

PR #705, opened 2017-05-12
Fix broken links in README.md

nacmartin:
Links to the Tutorial and Reference Documentation were broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,8 @@ console.log('hello, world!')
 
 # Learning More
 
-* Read the [tutorial](http://sweetjs.org/doc/1.0/tutorial.html).
-* Read the [reference documentation](http://sweetjs.org/doc/1.0/reference.html).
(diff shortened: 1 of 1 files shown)


gabejohnson:
Thanks for your contribution Nacho!

README.md
```diff
@@ -34,8 +34,8 @@ console.log('hello, world!')
 
 # Learning More
 
-* Read the [tutorial](http://sweetjs.org/doc/1.0/tutorial.html).
-* Read the [reference documentation](http://sweetjs.org/doc/1.0/reference.html).
+* Read the [tutorial](https://www.sweetjs.org/doc/tutorial.html).
+* Read the [reference documentation](https://www.sweetjs.org/doc/reference).
```
gabejohnson: Could you change the protocol on the editor link as well?

I don't think we need to remove the file extension. If we do, it should be done consistently.

@disnet?

disnet: I prefer no extension so removing it for tutorial would be good

PR #718, opened 2017-06-06
Do not include lang directive in output

disnet:
No reason for the `'lang sweet.js'` directive to be in the codegen'd output so omit it.

Also make all directives actually directives in the shift output AST so they don't get wrapped in parens.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
     ]
   },
   "files": [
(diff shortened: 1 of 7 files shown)


disnet:
@gabejohnson heh, glad you forced me to add tests. Definitely had a bug.

PR #723, opened 2017-06-11
Do not resolve exportedNames

### LINKED ISSUE #722: export should not `resolve`
e.g. `export var x` need to expand and codegen to `var x_1; export { x_1 as x }`

forms to check:

- `export var x`
- `export { x }`
- `export { x } from 'm'`

disnet:
Fixes #722.

Compiling `export var x` now compiles to `var x_1; export { x_1 as x }`.

### DIFF
--- a/src/sweet-module.js
+++ b/src/sweet-module.js
@@ -21,10 +21,20 @@ const extractDeclaration = _.cond([
 const ExpSpec = x => ({ exportedName: x });
 
 const extractDeclarationNames = _.cond([
-  [S.isVariableDeclarator, ({ binding }) => List.of(ExpSpec(binding.name))],
+  [S.isVariableDeclarator, ({ binding }) => List.of(binding.name)],
   [
     S.isVariableDeclaration,
     ({ declarators }) => declarators.flatMap(extractDeclarationNames)
+  ],
+  [S.isFunctionDeclaration, ({ name }) => List.of(name.name)],
+  [S.isClassDeclaration, ({ name }) => List.of(name.name)]
+]);
+
+const extractDeclarationSpecifiers = _.cond([
+  [S.isVariableDeclarator, ({ binding }) => List.of(ExpSpec(binding.name))],
+  [
+    S.isVariableDeclaration,
+    ({ declarators }) => declarators.flatMap(extractDeclarationSpecifiers)
   ],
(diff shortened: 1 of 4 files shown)


src/sweet-module.js
```diff
@@ -101,17 +111,59 @@ export default class SweetModule {
...
-        imports.push((item: any));
-      } else if (S.isExportDeclaration(item)) {
-        exports.push((item: any));
-        body.push(item);
-        this.exportedNames = this.exportedNames.concat(extractNames(item));
-        if (S.isExportDefault(item)) {
-          this.defaultExport = Syntax.fromIdentifier('_default');
-          this.exportedNames = this.exportedNames.push(
-            ExpSpec(this.defaultExport)
+      if (item instanceof T.ImportDeclaration) {
```
gabejohnson: Why aren't you using the spec helper predicates here?

test/parser/__snapshots__/test-ast.js.snap
```diff
@@ -69,7 +82,30 @@ Module {
...
+  "loc": null,
+  "type": "Module",
+}
+`;
+
+exports[`includes export with renaming in AST 1`] = `
+Module {
+  "directives": Array [],
+  "items": Array [
+    ExportFrom {
```
gabejohnson: Why are these `ExportFrom`?

src/sweet-module.js
```diff
@@ -101,17 +111,59 @@ export default class SweetModule {
...
+        imports.push(item);
+      } else if (item instanceof T.ExportDeclaration) {
+        if (S.isExport(item)) {
+          let decl = extractDeclaration(item);
+          let stmt = wrapStatement(decl);
+          let names = extractDeclarationNames(decl);
+          body.push(stmt);
+          let exp = new T.ExportFrom({
```
gabejohnson: Just answered my own question from the snapshot. We should add `ExportLocals` to sweet-spec.

src/sweet-module.js
disnet: because flow understands `instanceof`

src/sweet-module.js
disnet: That was added in shift spec es2016, we're on es6.

Definitely need to update but not in this PR.

PR #724, opened 2017-06-12
Fix too many invokes

### LINKED ISSUE #721: modules can be `invoke`d too many times
A module imported `for syntax` is currently re-`invoke`d for each module that imports it.

For example, given some modules like:

```js
// lib.js
export const x = 1;

// mod.js
import { x } from './lib' for syntax;

export syntax m = ctx => {
  x;
  return #`1`;
}

// main.js
import { x } from './lib' for syntax;
import { m } from './mod';

syntax n = ctx => {
  x;
  return #`2`;
}
```

we get an error:

```
evalmachine.<anonymous>:1
const x_1 = 1;
^

SyntaxError: Identifier 'x_1' has already been declared
```

/cc @gabejohnson I think this is the root cause of the error you mentioned the other day.

disnet:
fixes #721

### DIFF
--- a/src/multimap.js
+++ b/src/multimap.js
@@ -1,0 +1,21 @@
+// @flow
+import { List } from 'immutable';
+
+export default class Multimap<K, V> extends Map<K, List<V>> {
+  add(key: K, value: V) {
+    let bucket = this.get(key);
+    if (bucket != null) {
+      this.set(key, bucket.push(value));
+    } else {
+      this.set(key, List.of(value));
+    }
+  }
+
+  containsAt(key: K, value: V) {
+    let bucket = this.get(key);
+    if (bucket != null) {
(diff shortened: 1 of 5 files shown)


src/multimap.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+import { List } from 'immutable';
+
+export default class Multimap<K, V> extends Map<K, List<V>> {
+  add(key: K, value: V) {
+    let bucket = this.get(key);
+    if (bucket != null) {
+      this.set(key, bucket.concat(value));
```
gabejohnson: Should this be `bucket.push(value)`? `value` is `phase` (a number) in this instance. Correct?

disnet: ah, good catch

PR #734, opened 2017-07-16
Handle object destructuring with shorthand properties

### LINKED ISSUE #733: object destructuring shorthand broken for hygiene
```js
let { a } = b;
```

gets compiled to

```js
let { a_316 } = b;
```

but should instead be

```js
let { a: a_316 } = b;
```

This leads to some very confusing differences between using the shorthand notation and the longhand notation. 😠

disnet:
Fixes #733.

/cc @michaelficarra

### DIFF
--- a/src/enforester.js
+++ b/src/enforester.js
@@ -2,7 +2,7 @@
 import {
   isIdentifierExpression,
   isStaticMemberExpression,
-  isComputedMemberExpression,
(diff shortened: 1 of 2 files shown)


gabejohnson:
Is there an easy way to avoid including prettier format rule changes in these PRs? There's a lot of noise on these diffs.

Issue #662, opened 2017-03-30
Add module bundling


gabejohnson:
Implement the ideas in #633.

@disnet can you outline what you think the requirements are to make this happen?

PR #737, opened 2017-07-17
Handle forwarding bindings better

### LINKED ISSUE #736: importing names that have been "marked" by macro expansion are being renamed
Importing names that have "gone through a macro" causes renaming in the import clause:

```js
// mod.js
syntax m = ctx => {
  let first = ctx.next().value;
  let second = ctx.next().value;
  return #`class ${first} ${second}`;
}  
export m F { }

// main.js
import { F } from './mod.js';
let c = new F();
```

compiling `main.js` with sweet gives:

```js
import {F_5} from "./mod.js";
let c_7 = new F_5;
```

disnet:
fixes #736

Turns out there was an even simpler repro:

```js
// mod.js
export var x = 1;

// main.js
import { x } from './mod.js';
```

The issue is the design of module hygiene assumes bundling of all the modules #662 and so sets up a forwarding binding for each imported name. This greatly simplifies module bundling (just emit each module and bindings take care of themselves) but since we don't have bundling yet we run into this problem.

The solution is a little hacky, only add forwarding bindings to import names when:

- not the entrypoint or
- importing for phase > 0 or
- importing a compiletime binding

Don't merge just yet, I still need to add tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@sweet-js/core",
-  "version": "3.0.11",
+  "version": "3.0.12",
   "lockfileVersion": 1,
   "dependencies": {
     "@ava/babel-plugin-throws-helper": {
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "jquery": "3.1.1",
     "nyc": "^6.0.0",
     "prettier": "^1.5.3",
+    "rimraf": "^2.6.1",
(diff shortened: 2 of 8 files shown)


gabejohnson:
I see now that I couldn't repro because I wasn't compiling both files. I'll reserve further comment until the tests are added.

disnet:
Ok, tests are added

PR #741, opened 2017-08-19
Add parsing support for async/await

disnet:
This PR adds support in sweet's parser but shift-codegen doesn't support async/await yet so it won't be usable until support has been added there.

### DIFF
--- a/src/operators.js
+++ b/src/operators.js
@@ -8,6 +8,7 @@ const unaryOperators = {
   typeof: true,
   void: true,
   delete: true,
+  await: true,
 };
 const binaryOperatorPrecedence = {
   '*': 14,
--- a/src/enforester.js
+++ b/src/enforester.js
@@ -90,6 +90,7 @@ export class Enforester {
     phase: number | {},
     useScope: SymbolClass,
     bindings: any,
+    allowAwait?: boolean,
   };
   opCtx: {
     prec: number,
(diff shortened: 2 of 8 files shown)


src/operators.js
```diff
@@ -8,6 +8,7 @@ const unaryOperators = {
   typeof: true,
   void: true,
   delete: true,
+  await: true,
```
gabejohnson: It doesn't look like this is being used. If it does get used, is it correct? `await` is a contextual keyword like `yield` and `async`.

disnet: Yeah it is being used. The flow is a little non-obvious though. Basically adding `await` to `unaryOperators` allows [`enforestUnaryExpression`](https://github.com/sweet-js/sweet-core/blob/cdf3c03529057b4f8893fb6ffc066e40224bff2d/src/enforester.js#L2297) to parse `await` like the [unary operator](https://tc39.github.io/ecma262/#sec-unary-operators) it is. Later [`expandUnaryExpression`](https://github.com/sweet-js/sweet-core/blob/cdf3c03529057b4f8893fb6ffc066e40224bff2d/src/term-expander.js#L527-L531) converts `UnaryExpression`s with `await` to a proper AwaitExpression`.

Since sweet only parses modules `await` is not really a contextual keyword; it is always a keyword but only legal inside `async` functions. I not doing that validation but I suppose I probably should.

gabejohnson:
There's also an entry for `await` on line 84 of `tokens.js`

disnet:
Now enforcing that `await` is only valid in `async` contexts and all the left over references in tokens.js should be cleaned up.

